PR #15, opened 2017-07-02
add inventory model

### LINKED ISSUE #11: Create Inventory Model
This will be the model to encapsulate all the items into 1 group. This is the level where user permissions will be applied.
Need a name and description for now, we'll connect it to users later.

Exellin:
resolves #11

### DIFF
--- a/app/models/inventory.rb
+++ b/app/models/inventory.rb
@@ -1,0 +1,4 @@
+class Inventory < ApplicationRecord
+  validates :name, presence: true, length: { maximum: 255 }, uniqueness: { case_sensitive: false }
+  validates :description, length: { maximum: 255 }
+end
--- a/db/migrate/20170702185054_create_inventories.rb
+++ b/db/migrate/20170702185054_create_inventories.rb
@@ -1,0 +1,9 @@
+class CreateInventories < ActiveRecord::Migration[5.1]
+  def change
+    create_table :inventories do |t|
+      t.string :name, :limit => 255, :null => false
+      t.string :description, :limit => 255
+      t.timestamps
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -15,6 +15,13 @@ ActiveRecord::Schema.define(version: 20170703044736) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "inventories", force: :cascade do |t|
+    t.string "name", limit: 255, null: false
+    t.string "description", limit: 255
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
+
   create_table "users", force: :cascade do |t|
     t.string "email", default: "", null: false
     t.string "encrypted_password", default: "", null: false
@@ -26,10 +33,19 @@ ActiveRecord::Schema.define(version: 20170703044736) do
     t.datetime "last_sign_in_at"
     t.inet "current_sign_in_ip"
     t.inet "last_sign_in_ip"
+    t.string "confirmation_token"
+    t.datetime "confirmed_at"
+    t.datetime "confirmation_sent_at"
+    t.string "unconfirmed_email"
+    t.integer "failed_attempts", default: 0, null: false
+    t.string "unlock_token"
+    t.datetime "locked_at"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.index ["confirmation_token"], name: "index_users_on_confirmation_token", unique: true
     t.index ["email"], name: "index_users_on_email", unique: true
     t.index ["reset_password_token"], name: "index_users_on_reset_password_token", unique: true
+    t.index ["unlock_token"], name: "index_users_on_unlock_token", unique: true
   end
 
 end
--- a/spec/factories/inventory.rb
+++ b/spec/factories/inventory.rb
@@ -1,0 +1,8 @@
+require 'faker'
+
+FactoryGirl.define do
+  factory :inventory do
+    name { Faker::Name.unique.first_name }
+    description { Faker::Lorem.sentence }
+  end
+end
--- a/spec/models/inventory_spec.rb
+++ b/spec/models/inventory_spec.rb
@@ -1,0 +1,40 @@
+require 'rails_helper'
+
+RSpec.describe Inventory, type: :model do
+  it 'has a valid factory' do
+    expect(FactoryGirl.build(:inventory)).to be_valid
+  end
+
+  describe '#name' do
+    it 'is valid when under 256 characters' do
+      expect(FactoryGirl.build(:inventory, name: Faker::Lorem.characters(255))).to be_valid
+    end
+
+    it 'is invalid when nil' do
+      expect(FactoryGirl.build(:inventory, name: nil)).not_to be_valid
+    end
+
+    it 'is invalid when over 255 characters' do
+      expect(FactoryGirl.build(:inventory, name: Faker::Lorem.characters(256))).not_to be_valid
+    end
+
+    it 'is invalid with a duplicate value' do 
+      inventory = FactoryGirl.create(:inventory, name: 'duplicate')
+      expect(FactoryGirl.build(:inventory, name: 'Duplicate')).not_to be_valid
+    end
+  end
+
+  describe '#description' do
+    it 'is valid when nil' do
+      expect(FactoryGirl.build(:inventory, description: nil)).to be_valid
+    end
+
+    it 'is valid when under 256 characters' do
+      expect(FactoryGirl.build(:inventory, description: Faker::Lorem.characters(255))).to be_valid
+    end
+
+    it 'is invalid when over 255 characters' do
+      expect(FactoryGirl.build(:inventory, description: Faker::Lorem.characters(256))).not_to be_valid
+    end
+  end
+end


app/models/inventory.rb
```diff
@@ -0,0 +1,4 @@
+class Inventory < ApplicationRecord
+  validates :name, presence: true, length: { maximum: 255 }
+  validates :description, presence: true, length: { maximum: 255 }
+end
```
pachonk: New line at EOF please.

Exellin: done for all files

spec/models/inventory_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'rails_helper'
+
+RSpec.describe Inventory, type: :model do
+  it "has a valid factory" do
+    expect(FactoryGirl.build(:inventory)).to be_valid
+  end
+
+  describe "presence" do
+    it "is invalid with an empty name" do
+      inventory = Inventory.create(description: "this is a description")
```
pachonk: This test becomes less and less manageable as time moves on. I suggest doing something like this:
`expect(FactoryGirl.build(:inventory, name: nil)).not_to be_valid`

Exellin: updated all invalid tests to use factorygirl

spec/models/inventory_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe Inventory, type: :model do
+  it "has a valid factory" do
```
pachonk: Fair warning: this is me being nit picky. We will need to decide as a team if we want to follow this rule, but I follow it strictly personally.
Technically you should always use single quotes (`'`) unless you need string interpolation. They are faster to type (no need to press shift) as well as (marginally) faster for ruby to execute.
See: https://github.com/bbatsov/ruby-style-guide#strings

db/schema.rb
```diff
@@ -0,0 +1,23 @@
...
+# system, you should be using db:schema:load, not running all the migrations
+# from scratch. The latter is a flawed and unsustainable approach (the more migrations
+# you'll amass, the slower it'll run and the greater likelihood for issues).
+#
+# It's strongly recommended that you check this file into your version control system.
+
+ActiveRecord::Schema.define(version: 20170702185054) do
+
+  # These are extensions that must be enabled in order to support this database
+  enable_extension "plpgsql"
```
pachonk: (note: this is rails gen'd, so these quotes should never be manually changed)

spec/models/inventory_spec.rb
Exellin: Yeah I agree we should decide on a style for everything in the repo and follow it

Exellin: Fixed.

pachonk:
I have requested a review from @brandondees if he has time.
On any note, I'd like 1 more peer review before we pull this in.
Nice job man!

db/schema.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  create_table "inventories", force: :cascade do |t|
+    t.string "name", limit: 255
+    t.string "description", limit: 255
```
brandondees: if we're validating that these are absolutely required, that should be backed by a not-null constraint in the DB as well. in the real world it's quite common for some DB administrator to want to go manipulate data directly rather than through the application UI which can lead to apps breaking in annoying ways or even hard-to-troubleshoot ways. 

general rule of thumb: database constraints are consistency guarantees, rails validations are merely user guidance.

brandondees: (can be implemented as a separate issue / migration so as not to block acceptance of this PR, btw)

pachonk: Hey, good catch. I didn't catch that! I'm SUPER happy you reviewed for this very reason!

app/models/inventory.rb
```diff
@@ -0,0 +1,4 @@
+class Inventory < ApplicationRecord
+  validates :name, presence: true, length: { maximum: 255 }
+  validates :description, presence: true, length: { maximum: 255 }
```
brandondees: where are the user story requirements for these? i find in real world projects, description in addition to a name / primary identifier is very often considered unnecessary by users and they find it frustrating to require it. will the UI have an explanation for why these are essential, and how to fill them with useful content?

brandondees: @pachonk @Exellin ☝️

brandondees: should name also be unique for any reason? perhaps at least scoped within a given tenant / inventory?

brandondees: are these fields always meaningful to the user, and must they always be hand-chosen at creation time? does it make sense to have a safe default value generated so that users who don't care about these specifics can move along with their workflow more quickly? (I.e. randomly generate a unique name identifier at object instantiation to use as a #new form placeholder)

db/migrate/20170702185054_create_inventories.rb
```diff
@@ -0,0 +1,8 @@
+class CreateInventories < ActiveRecord::Migration[5.1]
+  def change
+    create_table :inventories do |t|
+      t.string :name, :limit => 255
+      t.string :description, :limit => 255
```
brandondees: no timestamps? or is that created implicitly now?

db/schema.rb
brandondees: there should probably be an issue to add a default .rubocop.yml and enforce linting for all contributions. rails-generated file locations can be whitelisted for exclusion from the rules

spec/models/inventory_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe Inventory, type: :model do
+  it 'has a valid factory' do
+    expect(FactoryGirl.build(:inventory)).to be_valid
+  end
+
+  describe 'presence' do
```
brandondees: is presence a method on the Inventory object? i think this could be structured / worded more clearly

spec/models/inventory_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    it 'is invalid with an empty name' do
+      expect(FactoryGirl.build(:inventory, name: nil)).not_to be_valid
+    end
+
+    it 'is invalid with an empty description' do
```
brandondees: it usually makes more sense to group assertions about a given attribute together under a `describe '#attr_name'` block rather than grouping them by common behavior, unless it really makes a big difference in terms of DRY code and is considered worth it.

brandondees: see http://www.betterspecs.org/ for a collection of generally good advice for rspec style. i disagree with some points but on average the rationale is clear enough and these rules make a decent baseline to follow by default

db/migrate/20170702185054_create_inventories.rb
pachonk: Ever since I've been using rails, timestamps are default.

app/models/inventory.rb
pachonk: I agree, let's not require presence of description, but let's require presence and uniq (later scoped to user/org) on name

spec/models/inventory_spec.rb
pachonk: Maybe we can group the 2 describe blocks and call em "validations"

brandondees: that's not my go-to pattern anymore but it works well enough.

db/migrate/20170702185054_create_inventories.rb
Exellin: added timestamps they weren't created by default with rails g migration.

spec/models/inventory_spec.rb
Exellin: I updated the specs to be grouped by attribute.  I'll also take a look through that link thanks for the resource.

app/models/inventory.rb
Exellin: updated name to be case sensitive and removed presence validation on description.

db/schema.rb
Exellin: So people go and update data on the database itself instead of through the rails console?  I was under the impression that ui validations were things handled on the browser level which won't let you submit a form until they are all met.  The validations in the model are still always present until you bypass both the ui and the rails console.

If this isn't the case, should I also research a way to validate that names are unique into the schema?

I added the not-null constrain to name and those constraints are automatically applied to timestamps in the meantime.

spec/factories/inventory.rb
```diff
@@ -0,0 +1,8 @@
+require 'faker'
+
+FactoryGirl.define do
+  factory :inventory do
+    name { Faker::Lorem.word }
```
pachonk: I love to use a "fun" faker method that has a lot of uniqueness. Also, use .uniq since we're validating uniqueness.

pachonk: https://github.com/stympy/faker#ensuring-unique-values

brandondees: uniq on faker means you'll run out of values if you test with too many instances. IMHO it's far safer to just concatenate the string with some true random data or use a sequence to include a non-repeating number as part of it

sanchez2210: Does this mean that the file should get modified now? (Adding random string)

pachonk: @sanchez2210 Yes, but let @Exellin do it since it's his PR :dancer:

pachonk: Or use a faker attribute that has MANY values.

spec/models/inventory_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    it 'is invalid with a duplicate value' do 
+      inventory = FactoryGirl.create(:inventory, name: 'duplicate')
+      expect(FactoryGirl.build(:inventory, name: 'Duplicate')).not_to be_valid
+    end
+  end
+
+  describe '#description' do
+    it 'is invalid when over 255 characters' do
```
pachonk: You should also explicitly test that it IS valid when it's nil and when it's < 255 char.

pachonk: Same for #name

Exellin: added those tests

spec/models/inventory_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  describe '#name' do
+    it 'is invalid when nil' do
+      expect(FactoryGirl.build(:inventory, name: nil)).not_to be_valid
+    end
+
+    it 'is invalid when over 255 characters' do
+      expect(FactoryGirl.build(:inventory, name: Faker::Lorem.characters(256))).not_to be_valid
+    end
+
+    it 'is valid when under 256 characters' do
```
pachonk: I typically like to group the valid and invalid together. You have a valid test right smack dab in the middle of these tests (same with desc, but with invalid)

Exellin: oh I was grouping it so the 256/255 value were right next to each other.

Exellin: done

spec/factories/inventory.rb
```diff
@@ -0,0 +1,8 @@
+require 'faker'
```
pachonk: This shouldn't need to exist since it's required from gemfile.

PR #37, opened 2017-07-03
Added .ruby-version file

### LINKED ISSUE #30: Add ruby version
I use `asdf` for both node and ruby, so I like `.tool-versions` but I'm guessing not everyone is.
We can either all move to `asdf`, which I don't expect everyone to do, or add the `.ruby-version` and the equal for node.

sanchez2210:
Resolves #30


pachonk:
Name of PR should match commit, description should contain the resolves piece.

pachonk:
You're also using an outlook email that isn't associated with your github account. Let's get that fixed and on the commit (or add your outlook to your github) before pulling this in.

sanchez2210:
Thank you for the reviews. I changed my github email since it was an old address. I'm not sure how to proceed now, should I push my commit again? Or create a new PR?

PR #52, opened 2017-07-04
Add rubocop_todo.yml to ignore rails generated files

### LINKED ISSUE #51: Configure rubocop to ignore rails generated files

pachonk:
I went back to the first commit and executed this:
bundle exec rubocop --auto-gen-config
echo 'inherit_from: .rubocop_todo.yml' > .rubocop.yml

Fixes #51


Exellin:
Do we need both files or could .rubocop.yml have all of the configuration instead of inheriting the todo file?

pachonk:
Ultimately, this config can shrink until it's things like only `schema.rb` as time moves on.

Exellin:
I just did a quick test and you don't need both files.

Also here is a quick list of files which we should add:

Guardfile
all files in db/migrate
db/schema.rb
spec/spec_helper.rb
spec/rails_helper.rb

Exellin:
Oh so the 2 files are so rubocop_todo is auto generated, and then you add your own exceptions in .rubocop.yml

pachonk:
Kind of.
The point is that we will over time fix the rails generated files at a natural pace and remove them from the rubocop todo.

pachonk:
So the _todo is LITERALLY "todo"

Exellin:
So we actually want to edit the rails generated files to adhere to rubocop standards over time?  Earlier you told me about some double quotation marks in some rails files which we shouldn't edit because they are generated by rails.

pachonk:
@Exellin Are you getting a lot of the frozen string literal error?

Exellin:
Yeah Missing magic comment # frozen_string_literal: true.

pachonk:
@Exellin Things like schema.rb need to be ignored because they're CONSISTENTLY auto gen'd.
Things like application.rb will eventually naturally conform to rubocop after time moves on.

pachonk:
@Exellin Then maybe we need to add a catch all for that rule in the TODO.
I gotta get going, will continue later.

Exellin:
Is there a reason that people who contribute to rails haven't already done that?  Or do the people who work on rubocop and people who work on rails disagree on standards?

pachonk:
A little bit of "idc" and a little bit of disagreements, probably. @brandondees explained all the reasons to me before but I forgot most of em.

brandondees:
rubocop is entirely independent and the default rubocop style rules are made up by @bbatsov based on opinions. most of those opinions are based on real rationalizations and overall the default rules are agreeable enough for a lot of the community, so it's a popular style guide. the rails project is governed by the rails core team and they have their own separate set of agreed upon style rules and I don't know whether they use an automated tool for enforcement or not. 

within any given team / project, it's up to the maintainers to configure their rubocop.yml to adjust the enforced rules to what the team can agree upon. some of the default rules may just be newer than the last time a given project or team agreed upon rules, and some of them may not make much sense for some projects.

brandondees:
there's also an element of "let's not change it all of a sudden just because of some style guide opinion since we've been doing it this way all along and it's consistent for us" -- understand that the rails core team doesn't care much what we freeloaders think of their opinions and practices

pachonk:
I need to spend around 15 minutes on this today, it's not completed. After it's completed, we can add to travis tests which will reject a PR that has code style errors :smile:

pachonk:
@Exellin Please proof. I just went ahead and mass added the rest of the issues in order to get this pulled in ASAP.
Will need to treat this like a real TODO now.

PR #53, opened 2017-07-05
Home page with link to sign in, sign up, sign out.

Exellin:
I had to make the sign out link a button because rails needs jquery_ujs to be able to make links other than a GET request.

I wasn't able to figure out how to specify the href on the button since buttons don't have href properties.

Also is specifying the href necessary?  I figured it would just add to readability to see exactly where the link is going in the test.

These links will eventually be moved to a navigation bar.

The links and button will eventually be assigned the same class so they won't differ in style.


spec/rails_helper.rb
```diff
@@ -6,7 +6,7 @@
 abort("The Rails environment is running in production mode!") if Rails.env.production?
 require 'rspec/rails'
 # Add additional requires below this line. Rails is not loaded until this point!
-
+include Warden::Test::Helpers
```
pachonk: Add a comment about why this piece is needed.

Exellin: done

app/controllers/pages_controller.rb
```diff
@@ -0,0 +1,4 @@
+class PagesController < ApplicationController
```
pachonk: Pages is rather generic. Can we brainstorm a better name for this (this might be the best name, I just like the general feel of the name "Page")

app/views/pages/home.html.erb
```diff
@@ -0,0 +1,6 @@
+<% if user_signed_in? %>
```
pachonk: Consider abstracting this to a partial.

spec/factories/users.rb
```diff
@@ -3,4 +3,8 @@
     email { Faker::Internet.email }
     password { Faker::Internet.password }
   end
+
+  factory :confirmed_user, parent: :user do
```
pachonk: Use a factory trait.

app/views/pages/home.html.erb
Exellin: What would that entail?  I thought partials were for reusing code between views.  This is already one line so I'm not sure exactly what that would entail and the purpose of doing so.

spec/features/pages/visit_root_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'rails_helper'
+
+RSpec.feature "Visiting root route" do
+  let(:user) { FactoryGirl.create :confirmed_user }
+
+  scenario "as a guest to the site shows sign up and sign in links" do
```
pachonk: Maybe add scenarios that the links go to devise's sign up and sign in as separate scenarios.

spec/factories/users.rb
Exellin: Like this?

  factory :confirmed_user, parent: :user do
    FactoryGirl.build(:user, confirmed_at: Time.now)
  end

It's giving me the missing host to link to error and it's not giving me a confirmed user.

app/controllers/pages_controller.rb
Exellin: I like pages.  I'm up for other suggestions.

spec/features/pages/visit_root_spec.rb
Exellin: Yeah I'm planning to create sign_in, sign_up, and sign_out features in a users/features folder.

spec/factories/users.rb
pachonk: https://robots.thoughtbot.com/remove-duplication-with-factorygirls-traits

spec/features/pages/visit_root_spec.rb
pachonk: Not ACTUALLY SIGN UP. Just that clicking the link takes you to the right place.

app/controllers/pages_controller.rb
pachonk: :+1:

pachonk:
Like issue this resolves in description.
Also, if it involves a view, css, js I recommend a screenshot too.

app/views/pages/home.html.erb
pachonk: Yes. This partial would then be rendered from the navigation directly.

spec/factories/users.rb
Exellin: done

app/views/pages/home.html.erb
Exellin: What is the purpose of creating a partial that is only referenced once (other than from application.html.erb to keep it clean).  Do we know in advance that the navbar is going to be so cluttered that we need to abstract away the authorization links, or is it convention to put items that need logic into a partial?

Exellin:
So I should reference the issue in both the commit message and the description of the pull?

I'll be sure to include screenshots in the future for any pages that have any amount of styling or layout, but this is literally just 2 links and a button.  Should I put in a screenshot anyway?

pachonk:
@Exellin I usually reference in the PR for certain, and sometimes in commit.

No need for this one, just figured I'd let you know!

app/views/pages/home.html.erb
pachonk: > other than from application.html.erb to keep it clean

Literally this reason

PR #59, opened 2017-07-05
Added the mailer url options to the test env config

sanchez2210:
The test environment cannot create users properly and we will need this for testing.


Exellin:
I also did this on my latest pull request.

sanchez2210:
Yes that wasn't live when I made this PR

PR #60, opened 2017-07-05
unnecessary variable removed

sanchez2210:


spec/models/inventory_spec.rb
```diff
@@ -18,8 +18,8 @@
       expect(FactoryGirl.build(:inventory, name: Faker::Lorem.characters(256))).not_to be_valid
     end
 
-    it 'is invalid with a duplicate value' do 
-      inventory = FactoryGirl.create(:inventory, name: 'duplicate')
+    it 'is invalid with a duplicate value' do
+      FactoryGirl.create(:inventory, name: 'duplicate')
```
pachonk: Name isn't an exact match.

sanchez2210: Yes I noticed that too but I think @Exellin did that to prove uniqueness: `{ case_sensitive: false }`?

Exellin: Yeah.  Although I guess it should maybe be in its own separate test?

pachonk: Let's do 2 different tests like so:
```ruby
describe 'uniqueness' do
  before do
    FactoryGirl.create(:inventory, name: 'Duplicate')
  end

  it 'something 1' do
    expect(FactoryGirl.build(:inventory, name: 'Duplicate')).not_to be_valid
  end

  it 'something2' do
    expect(FactoryGirl.build(:inventory, name: 'DuPliCaTe')).not_to be_valid
  end
end
```

pachonk: Can be a different PR. Just an idea.

Issue #43, opened 2017-07-03
How should we display form errors to users?


Exellin:
should we render a partial which renders all errors of the object above the form, create inline form errors, both, or another option which one of you can bring up?

pachonk:
I like doing an all flash partial.

Exellin:
I'm making this along with the authorization tests

PR #61, opened 2017-07-05
Create authorization feature specs

### LINKED ISSUE #54: Create sign_in, sign_up, and sign_out feature specs.
Edit:  I'm working on this right now.

Exellin:
resolves #54, #43

depends on #53 

I'll take any suggestions you have on color etc, and whether we want to change some of the devise flash messages to success and make a green alert as well.

the devise registrations/new view is auto generated except for the errors conditional and we will edit it further at a later date.

I added the javascript pack at the bottom of application.html.erb as I wasn't able to figure out how to set up document.ready without jquery if we want the javascript to be in the head of the application.

## Error messages on Sign Up Form
image

## Error message on Invalid Sign in
image

## Message on successful sign in
image

## Message on successful sign out
image

## Message on successful sign up
image

## Message on email confirmation
image

### DIFF
--- a/app/javascript/helpers/flashDismiss.js
+++ b/app/javascript/helpers/flashDismiss.js
@@ -1,0 +1,12 @@
+const flashDismiss = () => {
+  const removeAlert = (e) => {
+    document.body.removeChild(e.target.parentNode);
+  };
+  const dismissAlerts = document.getElementsByClassName('dismiss');
+
+  for (var dismiss of dismissAlerts) {
+    dismiss.onclick = removeAlert;
+  }
+};
+
+export default flashDismiss;
--- a/app/views/devise/registrations/new.html.erb
+++ b/app/views/devise/registrations/new.html.erb
@@ -1,0 +1,31 @@
+<h2>Sign up</h2>
+
+<% if resource.errors.any? %>
+  <%= render 'layouts/errors', obj: resource %>
+<% end %>
+
+<%= form_for(resource, as: resource_name, url: registration_path(resource_name)) do |f| %>
+  <div class="field">
+    <%= f.label :email %><br />
+    <%= f.email_field :email, autofocus: true %>
+  </div>
+
+  <div class="field">
+    <%= f.label :password %>
+    <% if @minimum_password_length %>
+    <em>(<%= @minimum_password_length %> characters minimum)</em>
+    <% end %><br />
+    <%= f.password_field :password, autocomplete: "off" %>
+  </div>
+
+  <div class="field">
+    <%= f.label :password_confirmation %><br />
+    <%= f.password_field :password_confirmation, autocomplete: "off" %>
+  </div>
+
+  <div class="actions">
+    <%= f.submit "Sign up" %>
+  </div>
+<% end %>
+
+<%= render "devise/shared/links" %>
--- a/spec/features/users/sign_in_spec.rb
+++ b/spec/features/users/sign_in_spec.rb
@@ -1,0 +1,22 @@
+require 'rails_helper'
+
+RSpec.feature 'Users signing in' do
+  let(:user) { FactoryGirl.create(:user, :confirmed) }
+
+  before do
+    visit '/'
+    click_link 'Sign In'
+  end
+  
+  scenario 'succeeds with valid credentials' do
+    fill_in 'Email', with: user.email
+    fill_in 'Password', with: user.password
+    click_button 'Log in'
+    expect(page).to have_content('Signed in successfully')
+  end
+  
+  scenario 'fails with invalid credentials' do
+    click_button 'Log in'
+    expect(page).to have_content('Invalid Email or password')
+  end
+end
--- a/spec/features/users/sign_up_spec.rb
+++ b/spec/features/users/sign_up_spec.rb
@@ -1,0 +1,23 @@
+require 'rails_helper'
+
+RSpec.feature 'Users signing up' do
+  let(:user) { FactoryGirl.build(:user) }
+
+  before do
+    visit '/'
+    click_link 'Sign Up'
+  end
+  
+  scenario 'succeeds with valid credentials' do
+    fill_in 'Email', with: user.email
+    fill_in 'Password', with: user.password
+    fill_in 'user_password_confirmation', with: user.password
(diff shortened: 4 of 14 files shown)


pachonk:
If a PR depends on another PR, put `depends on #issuenumber`

pachonk:
@snuggs could we get a js consult on this if you have any free time? It's only a few SLOC.

snuggs:
Sorry guys just missed this one. Was cleaning  out my notifications. @pachonk @Exellin

app/javascript/helpers/flashDismiss.js
```diff
@@ -0,0 +1,12 @@
+const flashDismiss = () => {
+  const removeAlert = (e) => {
+    document.body.removeChild(e.target.parentNode);
+  };
+  const dismissAlerts = document.getElementsByClassName('dismiss');
+
+  for (var dismiss of dismissAlerts) {
+    dismiss.onclick = removeAlert;
+  }
```
snuggs: Perhaps 

```javascript
document
  .getElementsByClassName ('dismiss')
  .map (dismiss => dismiss.onclick = removeAlert)

const
  removeAlert = e =>
    event.target.parentNode.remove ()
```

Much more [declarative vs imperative](https://tylermcginnis.com/imperative-vs-declarative-programming/).

app/views/devise/registrations/new.html.erb
```diff
@@ -0,0 +1,31 @@
+<h2>Sign up</h2>
+
+<% if resource.errors.any? %>
```
snuggs: Try to put the logic IN the helper.
```ruby
module SomeHelper
  def render_errors_for(resource)
    return unless resource.errors.any?

    render 'layouts/errors', obj: resource
  end
end
```
Then can just call `<%= render_errors_for resource %>` A #ProTip to "Logicless views" Doesn't mean there isn't any logic. Just means not in views.

app/views/devise/registrations/new.html.erb
```diff
@@ -0,0 +1,31 @@
...
+  <%= render 'layouts/errors', obj: resource %>
```
snuggs: `<%= render_errors_for resource %>`

Besides that this file is awesome 💯 👍

spec/features/users/sign_in_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails_helper'
+
```
snuggs: 💪  NICE!

spec/features/users/sign_in_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+RSpec.feature 'Users signing in' do
+  let(:user) { FactoryGirl.create(:user, :confirmed) }
```
snuggs: Might want to get used to `create` without `FactoryGirl`

https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#rspec

That said i'd use 
```ruby
let(:user) { attributes_for :user, :confirmed }
```

This way you aren't creating the user before adding a new one. Will blow up once you add `uniqueness_of` validator. I can already assume from this test the model doesn't have one.

spec/features/users/sign_up_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'rails_helper'
+
+RSpec.feature 'Users signing up' do
+  let(:user) { FactoryGirl.build(:user) }
+
+  before do
+    visit '/'
+    click_link 'Sign Up'
```
snuggs: seen this a few times. Might want to make a `spec_helper` function called `sign_up`
```ruby
before { sign_up }
```

spec/features/users/sign_in_spec.rb
pachonk: I actually don't like putting the build direct without the FG preface, personally

snuggs: Well whatever you do that's fine. But DEFINITELY use `attributes_for` in cases like this. For acceptance tests should NOT be able to directly touch database.

spec/features/users/sign_up_spec.rb
Exellin: That is just the way I have been testing my applications by making sure I can reach the page I am testing through the home page.  If it makes more sense I could directly visit new_user_registration_path.

If I continue this testing style, I would also visit '/' and click_link 'dashboard' then click_button 'New Organization' for creating an organization.  So this is not something unique to signing in or signing up that I think I need a helper function for.

app/views/devise/registrations/new.html.erb
Exellin: Thanks for the advice I'll be sure to export any logic to helper functions in the future and I'll refactor this one soon.

spec/features/users/sign_up_spec.rb
snuggs: Nah you wouldn't want to do that @Exellin because that is "Reaching into the black box" Again not knocking this pattern. Just helping conceptually move it forward. You did very well.

That said I GUARANTEE YOU when you test the "private" features of the app you're gonna want that `sign_up` helper in your before or no tests that require an account first will run.

ESPECIALLY when you change the text from `Sign Up` to `sign up` or something like that. You'd be painted in a corner having to make 1 change in `n` places within tests. Of course doesn't seem like a thought now. But tell that to your future self. :-)

Moreso about WHY you would do that. Always be in a situation that if a line of code has to change it only changes in 1 place (2 at the most).

My $0.02. YMMV

Exellin: In spec_rails_helper.rb I added 'include Warden::Test::Helpers'.

This has some helper methods for authentication and once we move on to private features of the application we will be using login_as(user).  You can see that in spec/features/users/sign_out_spec.rb.

However this still doesn't redirect capybara to a webpage, so I could add visit '/' to after login_as and clean up the tests that way if that makes sense.

snuggs: 100% @Exellin. It's worked for me a ton in the past but doesn't mean you shouldn't investigate doing what works for you.That's the difference between experience and advice (which i hate) hehe. Well done 💪

PR #66, opened 2017-07-06
Add Organization Model

### LINKED ISSUE #41: add organization model
## In Model
has a name attribute which is required, limited to a string of 255 characters, and is unique.

has_many inventories
HABTM users.

## In Schema
The name attribute will be not null and have a limit of 255.

Create table organization_users with organization id and user id with both indexed.

Needed for: #18 #19

DanBonehill:
Resolves #41 

Organization model added with HABTM association to User. Tests created and passed.

First ever PR... go easy on me :)


Exellin:
You need to rebase this branch on master so that is isn't comprised of 20 commits.

Also @pachonk I guess we will just deal with a possible merge conflict in development/test.rb for which ever PR is pulled in 2nd?

pachonk:
@Exellin Would you like to possibly walk him through rebasing and dropping those commits. Those commits are showing up because of something I did to repo way earlier.

pachonk:
I am spending around 15 minutes a day for guide writing, so I will have guides some day :)

DanBonehill:
Yeah I did wonder if I could do anything about that. I noticed that other PR's seem to be a few commits, whereas mine is quite a lot. I'll look up rebasing and see if I can work it out.

app/models/organization.rb
```diff
@@ -0,0 +1,5 @@
+class Organization < ApplicationRecord
+  validates :name, presence: true, uniqueness: true, length: { maximum: 255 }
```
Exellin: I think we should add case-senseitive: false on name otherwise you could have two organizations called Administrators and administrators.

spec/factories/organizations.rb
```diff
@@ -0,0 +1,10 @@
+FactoryGirl.define do
+  factory :organization do
+    name { Faker::Name.unique.first_name }
+    #association :users, :factory => :user
+
+    factory :organization_with_users do
```
Exellin: I feel like it is not necessary to create a separate factory for an organization with a user (sidepoint, this is an organization_with_user instead of organization_with_users.

If you want an organization with a user in a test you would just go Factorygirl.create(:organization, user: user)

Or since this has_many it might be Factorygirl.create(:organization, users: [user1, user2, user3]).  Can't remember the syntax off the top of my head.

This also requires you to uncomment the association.

spec/factories/users.rb
```diff
@@ -3,6 +3,10 @@
     email { Faker::Internet.email }
     password { Faker::Internet.password }
 
+    factory :users_with_organizations do
+      after(:build) {|user| user.organizations = [create(:organization)]}
+    end
+
```
Exellin: Same point here, I don't think we need a separate factory as you can just add any needed organizations on factory creation.

spec/models/organization_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      expect(FactoryGirl.build(:organization, name: 'duplicate')).not_to be_valid
+    end
+  end
+
+  describe '#user association' do
+    it 'is valid when it has users' do
+      expect(FactoryGirl.build(:organization_with_users)).to be_valid
+    end
+  end
+end
```
Exellin: I think it would be worthwhile to test whether the organizations are valid with 0, 1, or 2 users (right now you just have 0 and 1).  Also may need to add an additional test if we add case-sensitive: false.

app/models/organization.rb
pachonk: That brings another point. We should add an org blacklist, perhaps.

pachonk:
Closing in favor of new PR coming in. (on video call with him)

app/models/organization.rb
DanBonehill: Change Done, still on my branch

Issue #19, opened 2017-07-02
As a user, I want to be able to add an organization


pachonk:
That user should automatically be associated as the owner of the org.

dendenvitorio:
I'll work on this

dendenvitorio:
currently working on this

PR #68, opened 2017-07-06
Add organization model

### LINKED ISSUE #41: add organization model

DanBonehill:
Resolves #41

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -1,0 +1,5 @@
+class Organization < ApplicationRecord
+  validates :name, presence: true, uniqueness: true, length: { maximum: 255 }
+    has_many :inventories
+    has_and_belongs_to_many :users
+end
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -1,5 +1,5 @@
 class Organization < ApplicationRecord
-  validates :name, presence: true, uniqueness: true, length: { maximum: 255 }
-    has_many :inventories
-    has_and_belongs_to_many :users
+  validates :name, presence: true, uniqueness: { case_sensitive: false}, length: { maximum: 255 }
+  has_many :inventories
+  has_and_belongs_to_many :users
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -4,4 +4,5 @@ class User < ApplicationRecord
   devise :database_authenticatable, :registerable,
          :recoverable, :rememberable, :trackable, :validatable,
          :confirmable, :lockable
+  has_and_belongs_to_many :organizations
 end
--- a/db/migrate/20170704214210_create_organizations.rb
+++ b/db/migrate/20170704214210_create_organizations.rb
@@ -1,0 +1,9 @@
+class CreateOrganizations < ActiveRecord::Migration[5.1]
+  def change
+    create_table :organizations do |t|
+      t.string :name, :limit => 255, :null => false
+
+      t.timestamps
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170703044736) do
+ActiveRecord::Schema.define(version: 20170704214824) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -20,6 +20,19 @@ ActiveRecord::Schema.define(version: 20170703044736) do
     t.string "description", limit: 255
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+  end
+
+  create_table "organizations", force: :cascade do |t|
+    t.string "name", limit: 255, null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
+
+  create_table "organizations_users", id: false, force: :cascade do |t|
+    t.bigint "organization_id", null: false
+    t.bigint "user_id", null: false
+    t.index ["organization_id", "user_id"], name: "index_organizations_users_on_organization_id_and_user_id"
+    t.index ["user_id", "organization_id"], name: "index_organizations_users_on_user_id_and_organization_id"
   end
 
   create_table "users", force: :cascade do |t|
--- a/spec/factories/organizations.rb
+++ b/spec/factories/organizations.rb
@@ -1,0 +1,5 @@
+FactoryGirl.define do
+  factory :organization do
+    name { Faker::Name.unique.first_name }
+  end
+end
--- a/spec/models/organization_spec.rb
+++ b/spec/models/organization_spec.rb
@@ -18,9 +18,30 @@ RSpec.describe Organization, type: :model do
       expect(FactoryGirl.build(:organization, name: Faker::Lorem.characters(256))).not_to be_valid
     end
 
-    it 'is invalid with a duplicate organization' do
-      org = FactoryGirl.create(:organization, name: 'duplicate')
-      expect(FactoryGirl.build(:organization, name: 'duplicate')).not_to be_valid
+    context 'uniqueness' do
+      before do
+        FactoryGirl.create(:organization, name: 'duplicate')
+      end
+
+      it 'is invalid with a duplicate organization' do
+        expect(FactoryGirl.build(:organization, name: 'duplicate')).not_to be_valid
+      end
+
+      it 'is case insensitive' do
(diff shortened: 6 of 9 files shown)


Exellin:
Still requesting all changes made in #66

spec/models/organization_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      expect(FactoryGirl.build(:organization, name: Faker::Lorem.characters(256))).not_to be_valid
+    end
+
+    it 'is invalid with a duplicate organization' do
+      org = FactoryGirl.create(:organization, name: 'duplicate')
+      expect(FactoryGirl.build(:organization, name: 'duplicate')).not_to be_valid
+    end
+
+    it 'is case insensitive' do
+      org = FactoryGirl.create(:organization, name: 'Duplicate')
```
Exellin: you don't need to create the org variables here I think as you never use it.  @luis just made a commit recently deleting my unnecessary variable

DanBonehill: I thought the idea of the test was to create an Organization and then try and create another one with the same name but in a different case and expect that 2nd one not to be created?

Exellin: yeah factorygirl.create creates an organization in the database.  It doesn't have to be set to a variable.  factorygirl.build creates an object which isn't saved in the database yet.

DanBonehill: Oh ok, So I can remove `org =` and it'll still work?

Exellin: You can find that out by removing it and seeing if the tests still pass.

DanBonehill: 👍 Done

spec/models/organization_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    it 'is invalid when nil' do
+      expect(FactoryGirl.build(:organization, name: nil)).not_to be_valid
+    end
+
+    it 'is invalid when over 255 characters' do
+      expect(FactoryGirl.build(:organization, name: Faker::Lorem.characters(256))).not_to be_valid
+    end
+
+    it 'is invalid with a duplicate organization' do
+      org = FactoryGirl.create(:organization, name: 'duplicate')
```
Exellin: you forgot this one though.

spec/factories/organizations.rb
```diff
@@ -0,0 +1,6 @@
+FactoryGirl.define do
+  factory :organization do
+    name { Faker::Name.unique.first_name }
+
```
Exellin: don't need a blank line here if we are sticking with the style in spec/factories/inventories.rb

spec/models/organization_spec.rb
DanBonehill: Getting good practice at rebase here.

app/models/organization.rb
```diff
@@ -0,0 +1,5 @@
+class Organization < ApplicationRecord
+  validates :name, presence: true, uniqueness: { case_sensitive: false}, length: { maximum: 255 }
+    has_many :inventories
```
pachonk: Indent issue, sir.

app/models/user.rb
```diff
@@ -4,4 +4,5 @@ class User < ApplicationRecord
   devise :database_authenticatable, :registerable,
          :recoverable, :rememberable, :trackable, :validatable,
          :confirmable, :lockable
+  has_and_belongs_to_many :organizations
```
pachonk: Double space between line 6 and 7

db/migrate/20170704214210_create_organizations.rb
```diff
@@ -0,0 +1,9 @@
+class CreateOrganizations < ActiveRecord::Migration[5.1]
+  def change
+    create_table :organizations do |t|
+      t.string :name, :limit => 255, :null => false
+
+      t.timestamps
```
pachonk: t.timestamps should happen by default in Rails 5.

spec/models/organization_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    it 'is invalid with a duplicate organization' do
+      FactoryGirl.create(:organization, name: 'duplicate')
+      expect(FactoryGirl.build(:organization, name: 'duplicate')).not_to be_valid
+    end
+
+    it 'is case insensitive' do
+      FactoryGirl.create(:organization, name: 'Duplicate')
+      expect(FactoryGirl.build(:organization, name: 'duplicate')).not_to be_valid
+    end
+  end
```
Exellin: Alex commented on this in pull request #60 

It might be worthwhile to make a context block for uniqueness since we are repeating the line
FactoryGirl.create(:organization, name: 'Duplicate')
Something like

describe '#name'
  context 'uniqueness'
  
before do
  make the organization
end

it 'is invalid with a duplicate organization' do
make the test
end

it 'is case insensitive'  do
make the test
end
end

db/migrate/20170704214210_create_organizations.rb
Exellin: It doesn't.  I had to create timestamps in mine after in a new commit.

app/models/user.rb
DanBonehill: I don't get what you mean, should `has_and_belongs_to_many` be level will `:confirmable`?

db/migrate/20170704214210_create_organizations.rb
DanBonehill: That was there by default, I didn't add it

app/models/user.rb
Exellin: Think he just wants a blank line between the 2 properties of the model.

db/migrate/20170704214210_create_organizations.rb
Exellin: Yeah so rails might add timestamps to the migration by default, but if they are missing in the migration it won't create timestamps on the model.

app/models/user.rb
DanBonehill: I've made all changes except this one. Unsure exactly what you're asking here @pachonk

spec/models/user_spec.rb
```diff
@@ -14,4 +14,17 @@
...
+  describe '#organization association' do
+    it 'is valid when it has an organization' do
+      org = FactoryGirl.create(:organization)
+      expect(FactoryGirl.build(:user, organizations: [org])).to be_valid
+    end
+
+    it 'is valid when it has organizations' do
+      org1 = FactoryGirl.create(:organization)
+      org2 = FactoryGirl.create(:organization)
+      expect(FactoryGirl.build(:user, organizations: [org1, org2])).to be_valid
```
Exellin: You can create an array directly here.  

organizations = FactoryGirl.create_list(:organization, 2)

DanBonehill: Ah ok. Would it be better to not even create the variable. Just do:
`expect(FactoryGirl.build(:user, organizations: FactoryGirl.create_list(:organization, 2))).to be_valid`

Exellin: I think it is better to create the variable.  Creating a variable within another statement is not very readable if it is even possible.

DanBonehill: Ok 👍

app/models/user.rb
pachonk: Just a blank line to space these 2 things out a bit.

db/schema.rb
```diff
@@ -22,6 +22,19 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "organizations", force: :cascade do |t|
+    t.string "name", limit: 255, null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
+
+  create_table "organizations_users", id: false, force: :cascade do |t|
```
pachonk: We perhaps need a `owner` bool on here so we can generate routes like `pachonk/inventory-1` just like github.

Exellin: This will be resolved in #19.  What do you mean by bool though?  The organization will just belong_to an owner with a class of user.

pachonk: Think about it this way:

I'm the owner and I invite 2 friends to my inventory.
Now who's the owner?

Exellin: Still you.  why would inviting people change that?  We can implement a way to transfer ownership later.

pachonk: Sounds good

PR #71, opened 2017-07-07
Add css reset

### LINKED ISSUE #62: Add mobile viewport to application.html.erb
\<meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1">

This will set the width of the document equal to the width of the viewing device whether that is a desktop or phone.  If you look at the page now on a mobile device (you can do this on chrome by going into developer tools and clicking on the image with 2 devices left of elements) you'll notice that everything is zoomed out in order to fit everything on the page.

### LINKED ISSUE #63: Add CSS Reset
off the top of my head things to add

html, body {
  padding: 0;
  margin: 0;
}

Edit:  Also move css away from asset pipeline into webpacker.

Exellin:
resolves #62, #63 

Application.css wouldn't work so kept it as a scss file.
Moved css away from asset pipeline into webpacker.


public/packs-test/manifest.json
```diff
@@ -1,3 +1,4 @@
 {
+  "application.css": "/packs-test/application.css",
```
pachonk: Is this needed?

app/javascript/stylesheets/reset.scss
```diff
@@ -0,0 +1,4 @@
+html, body {
```
pachonk: It's not uncommon to reset more like this one: http://cssreset.com/scripts/eric-meyer-reset-css/
I'm cool with whatever, just figured I'd point it out.

.travis.yml
```diff
@@ -2,8 +2,12 @@ language: ruby
 cache: bundler
 rvm:
   - 2.4.1
+before_install:
+  - nvm install node
+  - npm install yarn
 before_script:
   - bin/setup
+  - bin/yarn
```
pachonk: I might be worth running bin/yarn inside of bin/setup so it does it on fresh clones too.

app/javascript/stylesheets/reset.scss
Exellin: I don't like adding css that is not needed and not a fan of adding things like no list style by default.

I guess CSS reset might be a bad name for what I was trying to here, it was just removing the margin on the body of the document.  Then if we ever want to add a reset or default in the future (which we need) we will add it to this file.

.travis.yml
Exellin: done

bin/setup
```diff
@@ -19,7 +19,7 @@ chdir APP_ROOT do
   system('bundle check') || system!('bundle install')
 
   # Install JavaScript dependencies if using Yarn
-  # system('bin/yarn')
+  system('bin/yarn')
```
pachonk: Haha, that was already in there default, just commented?!

Exellin: Yeah

PR #74, opened 2017-07-07
Improve inventory and authorization tests

Exellin:


Exellin:
Not sure how to reduce the lines in the inventory name block.

spec/models/inventory_spec.rb
```diff
@@ -18,9 +18,18 @@
       expect(FactoryGirl.build(:inventory, name: Faker::Lorem.characters(256))).not_to be_valid
     end
 
-    it 'is invalid with a duplicate value' do
-      FactoryGirl.create(:inventory, name: 'duplicate')
-      expect(FactoryGirl.build(:inventory, name: 'Duplicate')).not_to be_valid
+    context 'uniqueness' do
```
pachonk: More of a describe than a context. context is usually used to say "when" like "when a user is authenticated"

spec/features/users/sign_in_spec.rb
```diff
@@ -16,6 +16,13 @@
   end
   
   scenario 'fails with invalid credentials' do
+    fill_in 'Email', with: 'wrong_email@example.com'
+    fill_in 'Password', with: 'wrong_password'
+    click_button 'Log in'
+    expect(page).to have_content('Invalid Email or password')
```
pachonk: Rather than expecting it TO SHOW an error, you should expect it to NOT successfully sign in.

pachonk:
There's a broken rubocop rule currently. We should probably ignore that rule on spec/ dir in .rubocop.yml.

Exellin:
I'm going to fix and rebase this pull request once we fix and pull in the user and inventory relationship pull request which also has the same error.

pachonk:
@Exellin needs rebase for travis to pass.

Exellin:
There are still a few changes I want to make on this which involve the other PR so I'm holding off for now.

Exellin:
@pachonk updated and ready for another review.

PR #75, opened 2017-07-07
Import application.scss into application.js 

Exellin:
Webpack wasn't rendering application.js properly as it was already serving a pack with the same name

### DIFF
--- a/app/javascript/application.scss
+++ b/app/javascript/application.scss
@@ -1,0 +1,2 @@
+@import 'stylesheets/reset.scss';
+@import 'stylesheets/alerts.scss';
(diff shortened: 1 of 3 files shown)


pachonk:
Is there a reason that it doesn't work as application?

pachonk:
Also, might want to name "stylesheet" since it's a single stylesheet in the end.

Exellin:
If you check out the application right now, you can't dismiss the flash messages.  Webpack doesn't give an error or anything, but when I looked at the source js file and did a ctrl f for flashDismiss or alert nothing came up.

PR #76, opened 2017-07-08
Create navbar

Exellin:
Links are divied into logo, main links, and authorization spaced evenly on navbar

Below 768 pixels links are replaced with hamburger toggle to drop down all links

Added site-container class to container over yield with padding-top equal to nav width to keep all content visible

Let me know if you have any suggestions for color, layout etc.

## Desktop View
image

## Mobile collapsed
image

## Mobile expanded
image


pachonk:
I personally avoid collapsing menus at all cost when possible.

Exellin:
How do you fit all links into small screens?  Just have the links overflow into a new row?

pachonk:
@brandondees has some strong opinions on this matter as well.
Sometimes I do icons with really small font.

brandondees:
there are good arguments for and against a hamburger menu. fact is most folks who are tech-competent know what it is and how to use it, and fact is it's efficient in terms of screen space. if there's a reasonable way to avoid it, then that's also great. collapsing a row layout to use more vertical space is almost always a valid option (and generally is the default option) for doing usable responsive layout. design for the mobile experience first and then treat additional screen space as a progressive enhancement using media queries.

Exellin:
Updated to use the new stylesheet location.

app/views/layouts/_navbar.html.erb
```diff
@@ -0,0 +1,23 @@
+<nav>
+  <div class='container'>
```
brandondees: what problem does this extra div solve? always start with clean semantic markup and only add extra elements when they're functionally or semantically appropriate

app/views/layouts/_navbar.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    <div class='logo'>
```
brandondees: extraneous div?

app/views/layouts/_navbar.html.erb
```diff
@@ -0,0 +1,23 @@
...
+      <%= link_to 'Tenacious', root_path %>
+    </div>
+
+    <div id='nav-toggle'><span></span></div>
+
+    <!-- <div class='link-container'> -->
+    <!--  <%= link_to 'Nowhere', '#' %> -->
+    <!-- </div> -->
+
+    <div class='link-container'>
```
brandondees: why not use a list?

app/views/layouts/application.html.erb
```diff
@@ -9,8 +9,11 @@
   </head>
 
   <body>
-    <%= render 'layouts/flash_messages' %>  
-    <%= yield %>
+    <%= render 'layouts/navbar' %>
+    <div class='container site-container'>
```
brandondees: should this be `<main>`?

app/views/layouts/_navbar.html.erb
```diff
@@ -0,0 +1,23 @@
+<nav>
+  <div class='container'>
+    <div class='logo'>
+      <%= link_to 'Tenacious', root_path %>
+    </div>
+
+    <div id='nav-toggle'><span></span></div>
```
brandondees: what's the empty span for?

app/views/layouts/_navbar.html.erb
Exellin: The nav element is across 100% of the width of screen.  All of the links need to be encapsulated by another div or ul (if I am using li elements) with padding on either side to prevent the links from being at the edge of the screen.

app/javascript/stylesheets/navbar.scss
```diff
@@ -0,0 +1,103 @@
...
+  }
+}
+
+@media screen and (max-width: $breakpoint) {
+  #nav-toggle {
+    width: 35px;
+    height: 35px;
+    display: flex;
+    cursor: pointer;
+    span, span:before, span:after {
```
brandondees: i believe common style guides require a blank line before nested selectors... have we set up a scss linter yet?

app/views/layouts/_navbar.html.erb
Exellin: What is extraneous here?  I set up a logo class which would have a larger font size than the rest of the links on the navbar.  The link-container divs are also hidden when the navbar reaches the breakpoint so the only visible link left is the logo.

app/views/layouts/_navbar.html.erb
brandondees: I think the layout effect you want can be done pretty easily without the extra markup using flexbox, FWIW.

app/views/layouts/_navbar.html.erb
brandondees: just put the class on the link itself, no need for a wrapper div

pachonk:
Reminds me of my first front end review from @brandondees

brandondees:
it's pretty decent overall. i didn't look closely at the JS or CSS though.

app/views/layouts/_navbar.html.erb
Exellin: The span becomes the hamburger icon.

app/views/layouts/_navbar.html.erb
Exellin: I got rid of the container and switched space-between to switch-around to get rid of the needed padding on the edges.  It contains all of the links more towards the centre of the screen and I think it looks better.  Thanks for the advice.

image

app/views/layouts/_navbar.html.erb
Exellin: Any particular reason to use a list?  It just adds extra styling like list-style: none.

At the moment I am also grouping list items inside of a link-container div. like the following

nav
  a
  div.link-container
    a

If I switched to a list, I would have to go to something like the following.  If it is best practice or you can give me another reason I'll gladly switch.

nav
  ul
    li
      a
  ul.link-container
    li
      a

app/views/layouts/_navbar.html.erb
brandondees: ah, then it should have some kind of identifier indicating that. i would assume you can do the same thing with the div itself instead though

app/views/layouts/_navbar.html.erb
brandondees: a list is more semantically reflective of what is actually happening. you can nest lists and use them for dropdowns etc. also.

point isn't to reduce the amount of style rules needed, but to have the raw markup be as true a representation as possible of what it really means independently from styling. style rules are effectively free after the first page load, since they get cached if you're following normal practices, but html, particularly boilerplate shared by layouts, has to be loaded every time. the more semantically accurate it is, the more straightforward it is for both machines and humans to make sense of it and create new presentations for it. 

i try to think of it as if chrome and safari might start autonomously deleting my stylesheet and replace it with some general purpose one that's intended to work for arbitrary websites, and everything should continue to work and make sense without harming usability or losing functionality. in theory that should always be possible if you've kept your markup clean and correct, without too much expectation of inconsistencies or usability problems. that's just hypothetical though. the real benefits come from the clear separation of concerns between document and presentation which pays off longer term in terms of maintainability and often efficiency too.

app/views/layouts/_navbar.html.erb
brandondees: I think i'd do space-between and add some padding and margins as needed to get it looking sharp, but whatever floats your boat.

app/javascript/stylesheets/navbar.scss
Exellin: Nope we don't have one yet. I added spaces between each selector in the meantime.

app/views/layouts/application.html.erb
Exellin: That sounds reasonable.  Done.

pachonk:
We could make a custom <nav-element> html tagset :)

brandondees:
if it requires JS to work anyway, then might as well.

pachonk:
@Exellin I didn't notice the JS. This is actually doable in pure CSS btw!

Exellin:
Are there any benefits to a custom html tagset?  I understand their purpose in javascript frameworks for components but don't see any real benefits for something like a navbar.

Also should I research and implement this in pure CSS?  Not really sure where to start there.  I guess from a usability standpoint there might be a few users who don't have javascript enabled. The pure css solution might also be cleaner.

pachonk:
@Exellin https://www.baytreesolutions.com/ has a pure CSS navbar with a hamburger equal on mobile views, FWIW

brandondees:
pure css over JS is just a nice thing to do whenever feasible, namely for the sake of more universal accessibility with the resurgence in script blocking. in this case, we _know_ it's feasible because we just did it recently.

as for custom elements, there are pros and cons but ultimately for anything that needs JS to work in the first place, it's just a nicer way of packaging up that functionality into something clean and re-usable. it's exactly the same concept as using components / directives with JS frameworks, just no framework needed. that said, I have a serious preference for having everything work JS-free as much as possible. for this app it may not matter, as I'm not sure whether we can pull off some of the core functionality without using it, but it still can't hurt to practice the more ideal approach by default.

Exellin:
Pushed pure css solution

app/views/layouts/application.html.erb
```diff
@@ -3,14 +3,16 @@
...
-
     <%= stylesheet_pack_tag 'application' %>
     <meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1">
   </head>
 
   <body>
-    <%= render 'layouts/flash_messages' %>  
-    <%= yield %>
+    <%= render 'layouts/navbar' %>
+    <main class='container site-container'>
```
pachonk: Rather than adding these 2 classes, simply style the `main` tag DIRECTLY.

app/views/layouts/_navbar.html.erb
```diff
@@ -0,0 +1,22 @@
...
+
+  <label for='trigger' id='nav-toggle'>
+    <span></span>
+  </label>
+
+  <!-- <ul class='link-container'> -->
+  <!--  <li><%= link_to 'Nowhere', '#' %></li> -->
+  <!-- </ul> -->
+
+  <ul class='link-container'>
```
pachonk: `nav > ul`, no class

app/views/layouts/_navbar.html.erb
```diff
@@ -0,0 +1,22 @@
...
+    <% if user_signed_in? %>
+      <li><%= button_to 'Sign Out', destroy_user_session_path, method: :delete, class: 'link' %></li>
```
pachonk: `nav > ul button` instead of class.

brandondees: the `>` child selector should really only be used when the extra specificity of that direct relationship is necessary or semantically relevant. in this case maybe `body > nav button` instead of `nav > ul button` to target the "site main nav" as opposed to just any combination of ul inside of nav.

brandondees: (i know you got used to seeing this pattern in our recent project @pachonk  but i may have been over-using it there)

pachonk: Which pattern?

brandondees: the aggressive use of `>` selectors

pachonk: Ah, it was more to say "any direct ul of nav" but I get that.
Either way, we should go for FULL "markup that describes itself without classes" when possible.

app/views/layouts/application.html.erb
Exellin: Maybe I'm too used to bootstrap, but I really like the idea of a container class to be used throughout the site with padding on either side.  Do you have a better solution for this?

I can get rid of the site-container class and simply style the `nav ~ container` selector.  I can't applying padding to the `main` element (don't know exactly what rule is going on here) and if I apply a margin the margin of the first element in the main tag is ignored.

pachonk: You should be able to directly style the main tag, or even use `body > main` to style instead of classes.

Exellin: The padding is overwritten by .container and I can't find a way to be more specific and apply the styling.

app/views/layouts/_navbar.html.erb
Exellin: I used `form input` instead of `nav > ul button`.  

If we end up putting a second navbar into the site would it be frowned upon just to apply a class to one of them and style it off that?  There are already so many generic styles inside a nav inside this stylesheet that I feel it wouldn't be possible otherwise.

app/views/layouts/_navbar.html.erb
Exellin: done

app/views/layouts/application.html.erb
brandondees: `body > main` is very specific, so it should win in most specificity contests, I think. in the non-bootstrap world, `<body>` IS the layout-container element (`<html>` is the viewport backdrop, so to speak), and typical layouts will have a sandwich of `<header>`, `<main>`, `<footer>` inside of that, with a possible `<aside>` after `<main>` if you want a sidebar in the layout. everything else falls within one of those containers, and CSS takes care of how it looks visually.

I'm not sure I understand what problem you're having with removing the styling classes from `<main>` just apply the same style rules to `body > main` instead and the results should almost certainly be identical.

app/views/layouts/_navbar.html.erb
brandondees: Sounds like you're misunderstanding some of how CSS (and SCSS) work.

Your nested SCSS selector blocks generate additional nested selectors... 

```
nav {
  ul {
    a { display: block; }
  }
}
```

will render as 

```
nav ul a { display: block; }
```
which in this case is probably overly specific, as all nav / button links should probably be display: block by default. your `form input` example isn't "one less selector", it's one extra. specifying <form> as a parent is probably useless in this case, unless you specifically intend to use some inputs outside of form tags and rely on default styles for those. 

bear in mind that scss will generate separate output selector blocks for each combination of nested selectors, which can add a lot of weight to the total output if you're not crafting the rules mindfully. always be looking for opportunities to reduce the targeted specificity for rules that can be more generally applied. you can always override them as needed, and it generally reduces the need for duplicative rule declarations.

as for adding a styling class... I think you're still missing the point we're trying to help you understand: styling classes are almost always entirely unnecessary when using CSS effectively, and shouldn't be a default pattern except when using a framework that requires them in order to be used productively. one of the main benefits is to be able to use less markup and CSS (less page weight, faster load and render times) to achieve the same (or better) results. another main benefit is that you can make changes to presentation without being forced to update the templates to support it. styles just wrap around whatever the semantically-marked-up document is, rather than forcing the markup to bend out of shape to support the expectations of a stylesheet. once a lot of the basic defaults and structure are in place, it can wind up being a lot easier to maintain.

pachonk: That said, there definitely _are_ times you'll want to be this specific, IE:

You have 2 different anchors in the navbar and you want ones within the ul's to be block and anything non-ul to be inline-block, or something.

But if there's not multiple nests, you probably just want:
```scss
nav a {
  display: block;
}
```
or, if you really do want to be so specific
```scss
nav ul a {
  display: block;
}
```

app/javascript/stylesheets/navbar.scss
```diff
@@ -0,0 +1,136 @@
...
+  #nav-toggle {
+    display: none;
+  }
+  
+  nav > ul {
+    display: block !important;
+  }
+}
+
+.site-container {
```
pachonk: @Exellin This is the piece that should simply be `body > main`

Issue #31, opened 2017-07-03
Create administrators, either through a boolean in the devise user model or as another model entirely


Exellin:
I prefer to just apply a boolean to the user model and then surround each administrator action in an if statement checking if the user is an administrator.

pachonk:
@Exellin This is a little misleading. This makes it sounds like we want admins on the overall system, while I think we probably want to have admins on the level of repos and orgs. I propose that we hold off on "implementation issues" and stick to "user story" issues to make sure our implementation matches our stories requests.

Exellin:
Ok so we want an organization called "administators" with certain abilities.  My mistake on the misunderstanding.  Probably would be best to close the issue then if we are never going to implement administrators on this level.

pachonk:
Uhh, more like if I create an organization, see #17 
For now, assume that the client doesn't want there to be "super admins" that have control over the entire system until later.

Issue #29, opened 2017-07-03
set habtm relationship between user and inventory


Exellin:

pachonk:
Related to: #17

Exellin:
I created this issue as it needs to be implemented before we move on to #16 (which requires a controller and view) or 17 which you referenced already.  Or would it be better to create a pull request saying it as a partial implementation of those issues?

pachonk:
In that case, I put in this issue's description "Required for #16"

troystarwalt:
I'm working on this, but needed to add the defaul_url from #38 before finalizing the HABTM.

Exellin:
This can be closed now.

Exellin:
@troystarwalt are you working on this right now or still busy with family atm?

troystarwalt:
All yours!

Exellin:
This can be closed now.

Issue #79, opened 2017-07-09
As a user, I want to be able to add an inventory and associate it with myself or any of my organizations


pachonk:
Inventory.owner_id and Inventory.owner_type = [:user, :organization] which is therefor a "polymorphic" rails association could be an option.

Exellin:
I'm working on this right now

Exellin:
I assume this is only organizations you own, right?  You shouldn't be able to add any inventory to any organization you are just a part of.

pachonk:
I don't know, it could make sense someday to have either/or based on a flag.

Exellin:
o.k. I'll keep it as just owner for now.  Perhaps in the future we can put in a organization level like for inventories (read, write, admin).  If you have admin access you can create an inventory and associate it to the organization.

Exellin:
I tried to make an array with the current user and the owned inventories of the user for the owner of the inventory but wasn't able to get it to work.  Is that possible?

For the time being I am falling back to nested routes to grab the owner from the url.

pachonk:
You might need to make user.all_inventories that merges org invs and user.inventories (or something similar)

Exellin:
Don't think you understand my question.  User.inventories will already grab all inventories it owns (and is part of since we will add the user to inventory.users on create).

My issue is assigning an inventory an owner either through the form or on create.

Do we want to nest the inventories in this project?  For example if a user creates an inventory do we want it to live in /users/user_id/inventories/inventory_id.  Or just /inventories/inventory_id.  This is how Github works with repositories with either /organization or /user and I think it makes sense.

If that is the case, you can ignore my question since I was trying to get the owner functionality working in the /inventories/new route and wasn't sure how to pass in the owner in the form.

Issue #17, opened 2017-07-02
As an inventory administrator, I want to be able to associate a user to an inventory at the level or read, write, and admin


pachonk:

Exellin:
So If i understand what is going on here when a user creates an inventory, a relationship will be made between the two and the user will have access to read and write that inventory.

Then administrators will have the power to associate additional users to the inventory in order to give more users read and write access towards it.

pachonk:
admin would mean they have administrative rights over that "Inventory" they can change the name, desc, users allowed on repo, etc.

sanchez2210:
Do you think a HABTM is the way to go? I have some experience with them and came to the (maybe erroneous) conclusion that there's nothing a HABTM relationship will provide that a Has_many, Through: will not.

Exellin:
has_many, through: requires setting up an additional model and is preferred if you want to add any extra attributes to the relationship (such as a status etc).  HABTM is the same thing only slightly simpler to set up.  I'm not sure if it is best practice to always use has_many ,through: as it allows the flexibility of adding additional attributes in the future and accomplishes the same thing as HABTM.

sanchez2210:
Yes, that's my point. Also it's so much easier to add relationship validations on the join model. Specially at nested forms

pachonk:
I personally default to HABTM unless I need additional attributes, since it avoids setting up a model that's probably going to be mostly blank.
If I need to add additional attributes, that's when I convert it to a `has_many through:`

sanchez2210:
Okay, only that for new devs like me it's usually hard to know when exactly is recommended to turn it into a `has_many, through:` ( mostly As soon as some extra attributes on the relationship is needed ) and delay the switch too long ( I did, HM-T looked scary )

pachonk:
The simple answer is "when you need to have a model to back the intermediate table" the good news is you have 3 professional developers to help you make the call when it's ready to be converted to a has_many through.

sanchez2210:
Great, I will make the additions now.

sanchez2210:
#29

Exellin:
What additions are you making right now @sanchez2210?  @troystarwalt is already making the relationship in issue #29

sanchez2210:
Yeah, I checked through the issues and noticed that there was work already being done there.

Exellin:
I'm working on this at the same time as #29

Exellin:
[...]

PR #82, opened 2017-07-10
Create Inventory User association with user_roles

Exellin:
read, write, and admin levels through an enumerator
created inventory_users factory with the 3 roles as traits
tested inventory assocation in user_spec
tested user assocaition in inventory_spec

resolves: #31, #29

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -50,11 +50,6 @@ Layout/TrailingWhitespace:
     - 'spec/features/users/sign_up_spec.rb'
     - 'spec/rails_helper.rb'
 
-# Offense count: 3
-# Configuration parameters: CountComments, ExcludedMethods.
-Metrics/BlockLength:
-  Max: 35
-
 # Offense count: 65
 # Configuration parameters: AllowHeredoc, AllowURI, URISchemes, IgnoreCopDirectives, IgnoredPatterns.
 # URISchemes: http, https
@@ -66,21 +61,6 @@ Metrics/LineLength:
 Style/BlockComments:
   Exclude:
     - 'spec/spec_helper.rb'
-
-# Offense count: 9
-Style/Documentation:
-  Exclude:
-    - 'spec/**/*'
-    - 'test/**/*'
-    - 'app/controllers/application_controller.rb'
-    - 'app/controllers/pages_controller.rb'
-    - 'app/helpers/application_helper.rb'
-    - 'app/mailers/application_mailer.rb'
-    - 'app/models/application_record.rb'
-    - 'app/models/inventory.rb'
-    - 'app/models/organization.rb'
-    - 'app/models/user.rb'
-    - 'config/application.rb'
 
 # Offense count: 1
 # Cop supports --auto-correct.
--- a/app/models/inventory.rb
+++ b/app/models/inventory.rb
@@ -1,4 +1,7 @@
 class Inventory < ApplicationRecord
   validates :name, presence: true, length: { maximum: 255 }, uniqueness: { case_sensitive: false }
   validates :description, length: { maximum: 255 }
+
+  has_many :inventory_users
+  has_many :users, through: :inventory_users
 end
--- a/app/models/inventory_user.rb
+++ b/app/models/inventory_user.rb
@@ -1,0 +1,12 @@
+class InventoryUser < ApplicationRecord
+  belongs_to :user
+  belongs_to :inventory
+
+  enum user_role: {
+    'Read' => 0,
+    'Write' => 1,
+    'Admin' => 2
+  }
+
+  validates :user_role, inclusion: user_roles.keys
+end
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -1,5 +1,6 @@
 class Organization < ApplicationRecord
   validates :name, presence: true, uniqueness: { case_sensitive: false}, length: { maximum: 255 }
+
   has_many :inventories
   has_and_belongs_to_many :users
 end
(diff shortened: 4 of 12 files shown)


.rubocop_todo.yml
```diff
@@ -1,6 +1,6 @@
 # This configuration was generated by
```
pachonk: This file really shouldn't be touched unless there's a deletion after initial config. Why was it changed?

Exellin: We haven't talked too much about what rules we are going ignore or what styles we want to add to our models and classes etc to adhere to rubocop standards

I had the following errors before re-running the configuration file
spec/models/inventory_spec.rb:3:1: C: Block has too many lines. [45/35]
spec/models/user_spec.rb:3:1: C: Block has too many lines. [37/35]
app/models/inventory_user.rb:1:1: C: Missing top-level class documentation comment.

If you have some suggestions in how I can properly add another block to the tests I will gladly implement it but don't have any idea at the moment

Do we want to add a comment documenting what each class is for each ruby file?  If so I can start here.

Otherwise, I just added these offenses into the rubocop file since I didn't want to have this pull request fail because of rules/standards we have't decided on yet.

Is it better practice to bring in a pull request with failing rubocop tests that will be pulled in anyway, or how should I handle this in the future?

pachonk: TL;DR, busy, will read later.
Anything we don't want to enforce should go into the .rubocop.yml

Exellin: Do we want to enforce any of the above rules then or should I put them into .rubocop.yml?

pachonk: Let's blacklist spec for too many lines.
Let's do research on whether or not we want to enforce top level class comments as a team.

app/models/inventory_user.rb
```diff
@@ -0,0 +1,12 @@
+class InventoryUser < ApplicationRecord
+  belongs_to :user
+  belongs_to :inventory
+
+  enum user_role: {
```
pachonk: Isn't this really more of an inventory role than a user role?

spec/models/user_spec.rb
```diff
@@ -26,4 +26,23 @@
...
+    end
+
+    it 'is valid when it has an inventory' do
+      inventory = FactoryGirl.create(:inventory)
+      expect(FactoryGirl.build(:inventory_user, :read, inventory: inventory, user: @user)).to be_valid
+    end
+
+    it 'is valid when it has inventories' do
+      inventories = FactoryGirl.create_list(:inventory, 2)
+      (0..inventories.length - 1).each do |i|
```
pachonk: Why not just do `inventories.each do |inventory|`?

spec/models/inventory_spec.rb
```diff
@@ -37,4 +37,23 @@
       expect(FactoryGirl.build(:inventory, description: Faker::Lorem.characters(256))).not_to be_valid
     end
   end
+
+  describe '#user association' do
+    before do
```
pachonk: Use rspec `let`

pachonk: Or `subject`

app/models/inventory_user.rb
Exellin: It's a user's role on the inventory.  I think inventory role implies the role of the inventory (something like transport, sell, etc).

spec/models/inventory_spec.rb
Exellin: done

spec/models/user_spec.rb
Exellin: Just didn't think of it at the time.  Fixed for both files.

.rubocop_todo.yml
Exellin: I'm of the opinion that top level comments aren't really necessary for a rails projects unless you are adding a lot of custom functionality.  You know a controller is going to list available actions and fill in the logic for each action, and a model is going to list the relationships, validations, and custom models.

But I haven't really worked on a complicated rails projects and don't know how complicated classes can get and whether comments explaining the functionality of a class is a good habit to get in to even though it might feel useless to start off with.

Exellin:
I reverted the changes to rubocop_todo.yml and so far have the following in .rubocop.yml

Style/BlockLength:
  Exclude:
    - 'spec/**/*.rb'
    - Guardfile

then I also deleted the corresponding BlockLength in the todo file.

I also commented earlier on the comments for classes, what are your thoughts @pachonk?

app/models/inventory.rb
```diff
@@ -1,4 +1,7 @@
 class Inventory < ApplicationRecord
   validates :name, presence: true, length: { maximum: 255 }, uniqueness: { case_sensitive: false }
   validates :description, length: { maximum: 255 }
+
+  has_many :inventory_users
```
pachonk: Also need a has_one owner (of user or org)

Exellin: Yeah I assumed that was going to be implemented with #79

pachonk: Sounds good :ok_hand:

app/models/inventory_user.rb
pachonk: Can we use symbols instead of strings? `:read, :write, :admin`

app/models/organization.rb
```diff
@@ -1,5 +1,6 @@
 class Organization < ApplicationRecord
   validates :name, presence: true, uniqueness: { case_sensitive: false}, length: { maximum: 255 }
+
```
pachonk: Usually if you're not actually editing this file you'd want to leave an edit like this out for means of a more concise diff.

Exellin: @pachonk Would I make a separate PR for just this change, or leave it as is until I make a PR which changes this file?

app/models/inventory_user.rb
Exellin: @pachonk Since you already merged this I guess we can change this when we tackle issue #17

PR #87, opened 2017-07-12
Adding of organizations

dendenvitorio:
Resolves issue #19


Exellin:
General Comments before I get to individual files:  Try to avoid using scaffold to make anything in rails in the future.  There are a lot of files you made here which are not needed.

Also:  You should make a feature test for creating an organization.

app/assets/javascripts/organizations.js
```diff
@@ -0,0 +1,2 @@
+// Place all the behaviors and hooks related to the matching controller here.
+// All this logic will automatically be available in application.js.
```
Exellin: This file isn't needed.

app/assets/stylesheets/organizations.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the Organizations controller here.
+// They will automatically be included in application.css.
+// You can use Sass (SCSS) here: http://sass-lang.com/
```
Exellin: This file isn't needed.

app/helpers/organizations_helper.rb
```diff
@@ -0,0 +1,2 @@
+module OrganizationsHelper
+end
```
Exellin: This file isn't needed until you want to create a helper method.

app/views/layouts/application.html.erb
```diff
@@ -11,5 +11,11 @@
     <%= render 'layouts/flash_messages' %>  
     <%= yield %>
     <%= javascript_pack_tag 'application' %>
+    
+<% if user_signed_in? %>
+  <%= link_to "Add Organization", new_organization_path %> 
+<% end %>
+
+
```
Exellin: You shouldn't put any logic or links directly into application.html.erb.  You should put this into home.html.erb or even create a user authenticated dashboard in the pages controller.  Then you don't need to check if the user is signed in to add the link.

app/views/organizations/_form.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= form_for @organization, url: {action: "create"}, html: {class: "nifty_form"} do |f| %>
+  <%= f.text_field :name %>
+  <br>
+  <%= f.submit "Create" %>
+<% end %>
```
Exellin: Not too worried about the actual styling or layout of the form right now, but you don't have a nifty_form class so you shouldn't add it here.  Also no newline at end of file.

config/locales/simple_form.en.yml
```diff
@@ -0,0 +1,31 @@
...
+    # hints:
+    #   defaults:
+    #     username: 'User name to sign in.'
+    #     password: 'No special characters, please.'
+    # include_blanks:
+    #   defaults:
+    #     age: 'Rather not say'
+    # prompts:
+    #   defaults:
+    #     age: 'Select your age'
```
Exellin: This file isn't needed since we aren't using simple form.

db/schema.rb
```diff
@@ -26,6 +26,7 @@
     t.string "name", limit: 255, null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.integer "user_id"
```
Exellin: I might be missing something here, but is there no migration for this?  Did you hard code the integer inside?  Never ever touch the schema file since I believe that won't actually alter the database and it will just cause confusion.

lib/templates/erb/scaffold/_form.html.erb
```diff
@@ -0,0 +1,13 @@
...
+  <div class="form-inputs">
+  <%- attributes.each do |attribute| -%>
+    <%%= f.<%= attribute.reference? ? :association : :input %> :<%= attribute.name %> %>
+  <%- end -%>
+  </div>
+
+  <div class="form-actions">
+    <%%= f.button :submit %>
+  </div>
+<%% end %>
```
Exellin: don't need this file since we aren't using simple form.

spec/controllers/organizations_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe OrganizationsController, type: :controller do
+
+end
```
Exellin: Don't need this file as we aren't doing controller specs.  Everything that we could test in a controller we are testing in feature specs.

spec/helpers/organizations_helper_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+# describe OrganizationsHelper do
+#   describe "string concat" do
+#     it "concats two strings with spaces" do
+#       expect(helper.concat_strings("this","that")).to eq("this that")
+#     end
+#   end
+# end
+RSpec.describe OrganizationsHelper, type: :helper do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
```
Exellin: Don't need this file.

app/controllers/organizations_controller.rb
```diff
@@ -0,0 +1,24 @@
+class OrganizationsController < ApplicationController
+	before_action :authenticate_user!, only: [:new]
```
Exellin: you should also authenticate the create action.  If I understand it correctly  if someone bypasses the meta tokens and submits a post request to make a new organization they would be able to do so.

app/controllers/organizations_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+
+	def index
+		@organization = Organization.all
```
Exellin: The story we are working with right now never specified that a user wants to go to a page to see all organizations they are a part of.  We might implement this into a dashboard in the future.

app/controllers/organizations_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+	end
+
+	def new 
+		@organization = current_user.organizations.build
+		@users = User.all.map{ |u| [u.id] }
```
Exellin: What purpose is building organizations and creating a map of all users doing here?

app/controllers/organizations_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+	end
+
+	 def create
+        @organization = current_user.organizations.build(org_params) 
+        current_user.organizations << @organization
+        @organization.save
```
Exellin: I'm just asking again here what is the purpose of building instead of creating a new model off of the class.
Also you don't need to manually add an organization into the organizations array.  A database call will create that array for us as long as the join table is set up properly.

app/controllers/organizations_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+	private
+
+	def org_params
+		params.require(:organization).permit(:name, :user_ids => [])
```
Exellin: There is only one user id on an organisation so there shouldn't need to be an array here.

PR #91, opened 2017-07-13
Add create organization feature

### LINKED ISSUE #19: As a user, I want to be able to add an organization

Exellin:
Resolves #19 

My spec focused on separating the creation of the organization with the ui messages showing an organization was created or if there were errors.

I didn't test for every way the form could fail as I feel that would be a repeat of all the validations in the model spec

I also asserted things like expect(current_path).to eq root_path in the spec which will have to change when we redirect to a more meaningful route.

Added a success alert with a green background

The link for adding an organization will eventually be moved to a profile or dashboard page wherever we decide in the future

I left the form unstyled for now as I am not very familiar with styling forms but we can create an issue for users wanting visually appealing forms in the future

Rendered out errors_for into application helper

Duplicated the rubocop configuration from #82 so it shouldn't have a merge conflict for whichever of these are pulled in first

I didn't create a form partial as we haven't decided on a user story for editing an organization yet.

### DIFF
--- a/app/javascript/stylesheets/alerts.scss
+++ b/app/javascript/stylesheets/alerts.scss
@@ -5,7 +5,7 @@
   justify-content: space-between;
 
   span {
-    align-self: center
+    align-self: center;
   }
 
   .dismiss {
@@ -16,9 +16,13 @@
 }
 
 .alert-alert {
-  background-color: #da1717
+  background-color: #da1717;
 }
 
 .alert-notice {
-  background-color: #3691ca
+  background-color: #3691ca;
+}
+
+.alert-success {
+  background-color: #5a9e21;
 }
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,5 +1,5 @@
 Rails.application.routes.draw do
   root 'pages#home'
   devise_for :users
-  # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  resources :organizations, only: [:new, :create]
 end
--- a/spec/features/organizations/create_organization_spec.rb
+++ b/spec/features/organizations/create_organization_spec.rb
@@ -1,0 +1,58 @@
+require 'rails_helper'
+
+RSpec.feature 'Creating an organization' do
+  let(:user) { FactoryGirl.create(:user, :confirmed) }
+  let(:organization) { FactoryGirl.build(:organization, owner: user) }
+
+  before do
+    visit '/'
+  end
+
+  feature 'as an authenticated user' do
+    let!(:original_organization_count) { Organization.count }
+
+    before do
+      login_as(user)
+      click_link 'Create an Organization'
+    end
+
+    feature 'with valid inputs' do
+      before do
+        fill_in 'organization_name', with: organization.name
+        click_button 'Create Organization'
+      end
+
+      scenario 'succeeds' do
+        created_organization = Organization.last
+        expect(Organization.count - original_organization_count).to be(1)
+        expect(created_organization.owner).to eq(organization.owner)
+        expect(created_organization.name).to eq(organization.name)
+        expect(current_path).to eq(root_path)
+      end
+
+      scenario 'shows a message saying the organization was created' do
+        expect(page).to have_content('Your organization has been successfully created')
+      end
+    end
+
+    feature 'with no inputs' do
+      before do
+        click_button 'Create Organization'
+      end
+
+      scenario 'fails' do
+        expect(Organization.count).to eq(original_organization_count)
+      end
+
+      scenario 'shows messages saying why the organization was not created' do
+        expect(page).to have_content("Name can't be blank")
+      end
+    end
+  end
+
+  scenario 'as a guest redirects to the log in page' do
+    click_link 'Create an Organization'
+    expect(page).to have_content('You need to sign in or sign up before continuing.')
+    expect(current_path).to eq(new_user_session_path)
+  end
+end
--- a/spec/models/organization_spec.rb
+++ b/spec/models/organization_spec.rb
@@ -33,15 +33,22 @@ RSpec.describe Organization, type: :model do
     end
   end
 
-  describe '#user association' do
-    it 'is valid when it has a user' do
-      user = FactoryGirl.create(:user)
-      expect(FactoryGirl.build(:organization, users: [user])).to be_valid
+  describe '#owner' do
+    it 'is invalid when nil' do
+      expect(FactoryGirl.build(:organization, owner: nil)).not_to be_valid
     end
 
-    it 'is valid when it has users' do
-      users = FactoryGirl.create_list(:user, 2)
-      expect(FactoryGirl.build(:organization, users: users)).to be_valid
+    it 'is a user' do
+      organization = FactoryGirl.create(:organization)
+      expect(organization.owner).to be_a(User)
+    end
+  end
+
+  describe '#users' do
+    it 'returns users belonging to the organization' do
+      organization = FactoryGirl.create(:organization)
+      users = FactoryGirl.create_list(:user, 2, organizations: [organization])
+      expect(organization.users).to eq(users)
     end
   end
 end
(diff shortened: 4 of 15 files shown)


bin/setup
```diff
@@ -28,7 +28,7 @@ chdir APP_ROOT do
   # end
 
   puts "\n== Preparing database =="
-  system! 'bin/rails db:setup'
+  system! 'bin/rails db:reset'
```
pachonk: Ah, no this shouldn't be changed. Sorry for being unclear.

spec/features/organizations/create_organization_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+  let(:user) { FactoryGirl.create(:user, :confirmed) }
+  let(:organization) { FactoryGirl.build(:organization, owner: user) }
+
+  before do
+    visit '/'
+  end
+
+  feature 'as an authenticated user' do
+    before do
+      @original_organization_count = Organization.count
```
pachonk: I usually will abstract this to a `let`

spec/models/user_spec.rb
```diff
@@ -26,4 +26,18 @@
       expect(FactoryGirl.build(:user, organizations: orgs)).to be_valid
     end
   end
+
+  describe '#owned organizations' do
```
pachonk: What if it owns 0 orgs? Is it still valid?

spec/models/organization_spec.rb
```diff
@@ -33,6 +33,17 @@
     end
   end
 
+  describe '#owner' do
+    it 'is invalid when nil' do
+      expect(FactoryGirl.build(:organization, owner: nil)).not_to be_valid
+    end
+
+    it 'is a user' do
```
pachonk: Did you try to see if the spec description read nicely without the `'is a user'? Sometimes the descriptions that the matchers have are awesome and it's nice to use them when possible because they will always be accurate to what you're asserting.

config/routes.rb
```diff
@@ -1,5 +1,6 @@
 Rails.application.routes.draw do
   root 'pages#home'
   devise_for :users
+  resources :organizations, only: [:new, :create]
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
```
pachonk: Can we delete the comment too?

app/javascript/stylesheets/alerts.scss
```diff
@@ -16,9 +16,13 @@
 }
 
 .alert-alert {
-  background-color: #da1717
+  background-color: #da1717;
```
pachonk: Maybe we should abstract the red, green, etc out to global vars so we can change them whenever we want. Thoughts?

bin/setup
Exellin: So I should just rails db:reset before I do bin/setup when aligning the database with the current branch?

spec/models/user_spec.rb
Exellin: It seems very strange imo to test whether the parent of a has_many relationship is valid without any of the child objects unless we added some sort of after_create method on a user which automatically created a organization for it.  All of our other tests checking to see if something was valid when nil or 0 was when that object had an attribute which we could set to nil in the database.

In the organization_spec I built a test seeing if an organization was invalid without a user with the following: 
`expect(FactoryGirl.build(:organization, owner: nil)).not_to be_valid`

How would I test if a user is valid if it has 0 organizations?  I can't go 
`expect(FactoryGirl.build(:user, owned_oranizations: nil)).to be_valid`

Since owned_organizations is essentially a method which searches the organizations table with organizations which have an owner id equal to the id of the user the method was called on and returns an ActiveRecord Collection.

Since we are creating a user at the beginning of this file and there are no organizations in the user factory that implies it is valid with 0 owned organizations.

Also was it fine that we didn't test if the user was valid with 0 organization associations?

I just came up with this test:

```
    it 'is valid when it has no organizations' do
      expect(user.owned_organizations.empty?).to be true
    end
```

Since owned_organizations returns an activeRecord collection I can't test to see if it is equal to nil, to test if it is valid, or to test if it is an empty array so this is the only test I could find which made some sort of sense, but I feel it is not needed and redundant.

pachonk: Hmm, I'd think that `expect(FactoryGirl.build(:user, owned_oranizations: nil)).to be_valid` would work.

Exellin: I don't see how that would work.  factories work by applying a value to an attribute of the object.  owned_organizations isn't an attribute.  It is a method which looks for owned organizations of the user.

This is the error I get when using 
`expect(FactoryGirl.build(:user, owned_organizations: nil)).to be_valid`

`     NoMethodError:
       undefined method `each' for nil:NilClass
`

app/javascript/stylesheets/alerts.scss
Exellin: Why can't we change them whenever we want with them still in the file?

spec/models/user_spec.rb
pachonk: Try `expect(FactoryGirl.build(:user, owned_oranizations: [])).to be_valid`

spec/models/organization_spec.rb
Exellin: just call the test it do?  

I just did a quick test by making the test fail and it printed the following

```
Organization#owner should not be a kind of User(id: integer, email: string, encrypted_password: string, reset_password_token: string, reset_pass...pts: integer, unlock_token: string, locked_at: datetime, created_at: datetime, updated_at: datetime)
```

So you can still decipher it when the test fails, but then when someone is reading over the test how do they know what it is testing without looking at the code inside the test?

pachonk: It looks pretty clear it's saying the same thing and it's self documenting of what you're asserting too. It's not the best matcher description there is, but it's not a terrible one. I always consider if I can omit that description arg on an `it` and sometimes it's worth it, sometimes it's not. If it is worth it, it's preferred because it will change if the assertion change.

spec/models/user_spec.rb
Exellin: Yeah that works but it still seems very unnecessary as I think a test like this would be done on the actual ActiveRecord class (that an active record collection can be empty)

If you still want me to put it in I will.  Also should I do the same thing with the organization association (test that it can have 0 associations?)

pachonk: It just seemed like something you'd want to test as well if you're testing that it's valid WITH orgs.

spec/models/organization_spec.rb
Exellin: I really enjoying reading what a test is testing without having to go through the actual test line by line.  If you change what a test is doing it seems reasonable that you would also have to change the description of the test, and you get to keep the readability.

pachonk: To me, it reads the same, or at least similar enough that I'd go with the "less manual description", but hey it's your PR. Maybe we can get more insight too.
`owner should not be a kind of User` (I assume you made it to_not to make it fail) it would directly describe EXACTLY what's happening.

spec/models/user_spec.rb
Exellin: Maybe I'm not naming the test correctly then.  It's more that I am testing the connection between organizations and owners.  Maybe the word validation should be kept for attributes in a database.

In organization_spec I am making sure the owner can't be nil, and that when I create an owner `organzation.owner` will return the correct user

In user_spec I want to make sure that when an organization is made with an owner, `user.organizations` will return the organizations with owner_id equal to user_id.

pachonk: Yeah, then maybe it's a test misnomer.

spec/models/organization_spec.rb
Exellin: Where are you putting `owner should not be a kind of User`.  I thought you wanted to change it to it do.  With no description which is what I don't want to implement since the only description of the test is in the code itself.

pachonk:
@Exellin conflicts.

Issue #93, opened 2017-07-19
As a user, I would like to see a list of organizations I am a part of on my dashboard


pachonk:

Exellin:
I'll work on this right now.

PR #94, opened 2017-07-20
Create dashboard with list of organizations and organization show action

Exellin:
Added current_user to organization.users on create
Moved Create organization link from homepage to dashboard
Created will_redirect_to_login_page helper method for feature specs
Create show_organization_spec and visit_dashboard_specs specs

resolves: #93

### DIFF
--- a/app/views/pages/dashboard.html.erb
+++ b/app/views/pages/dashboard.html.erb
@@ -1,0 +1,7 @@
+<% @organizations.each do |organization| %>
+  <ul>
+    <li><%= link_to organization.name, organization_path(organization) %></li>
+  </ul>
+<% end %>
+
+<%= link_to 'Create an Organization', new_organization_path %>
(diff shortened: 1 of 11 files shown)


app/views/pages/dashboard.html.erb
```diff
@@ -0,0 +1,7 @@
+<% @organizations.each do |organization| %>
```
pachonk: Why not just use `current_user.organizations.each`?

app/views/pages/dashboard.html.erb
```diff
@@ -0,0 +1,7 @@
...
+  <ul>
+    <li><%= link_to organization.name.to_s, organization_path(organization) %></li>
```
pachonk: Shouldn't name ALWAYS return a string?

Exellin: Yeah forget why I put that in there, good catch.

app/views/pages/dashboard.html.erb
Exellin: Just less logic in the view so I could put @organizations = current_user.organizations into the pages_controller.

Or do you only put instance variables in the controller to avoid calling class methods in views like Organization.all?

PR #95, opened 2017-07-23
updates password min-length to 14 and updates factory

### LINKED ISSUE #72: As a developer, I was to ensure that users have secure passwords
No symbol restrictions and min length 14 is my recommendation.

troystarwalt:
Resolves #72

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -22,6 +22,13 @@ ActiveRecord::Schema.define(version: 20170713185103) do
     t.datetime "updated_at", null: false
   end
 
+  create_table "inventories_users", id: false, force: :cascade do |t|
+    t.bigint "inventory_id"
+    t.bigint "user_id"
+    t.index ["inventory_id"], name: "index_inventories_users_on_inventory_id"
+    t.index ["user_id"], name: "index_inventories_users_on_user_id"
+  end
+
   create_table "inventory_users", force: :cascade do |t|
     t.bigint "user_id", null: false
     t.bigint "inventory_id", null: false
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -157,7 +157,7 @@ Devise.setup do |config|
 
   # ==> Configuration for :validatable
   # Range for password length.
-  config.password_length = 6..128
+  config.password_length = 14..128
 
   # Email regex used to validate email formats. It simply asserts that
   # one (and only one) @ exists in the given string. This is mainly
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -1,7 +1,7 @@
 FactoryGirl.define do
   factory :user do
     email { Faker::Internet.email }
-    password { Faker::Internet.password }
+    password { Faker::Internet.password(14, 128) }
 
     trait :confirmed do
       confirmed_at Time.now


db/schema.rb
```diff
@@ -22,6 +22,13 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "inventories_users", id: false, force: :cascade do |t|
+    t.bigint "inventory_id"
+    t.bigint "user_id"
+    t.index ["inventory_id"], name: "index_inventories_users_on_inventory_id"
+    t.index ["user_id"], name: "index_inventories_users_on_user_id"
+  end
+
```
Exellin: why is there a second inventories_users table here?

Exellin:
Might want to add a unit test for password length.

db/schema.rb
troystarwalt: Hmm that must be from an old merge of mine. Didn't catch that.

troystarwalt:
Would love a tip to get started on how to test on models. I thought the current coverage would work as you can't create a user, at all, without the password being 14 characters or more.

troystarwalt:
Would something like this work?

>   describe User do
>    it "validates password length" do
>      FactoryGirl.build(:user, password: "1234567").should_not be_valid
>      FactoryGirl.build(:user, password: "a42d1edd2637268dbd053521828ed3eecd5e9e2e95bea57a14d5fe453ba1399c66854004a3d7e13b16d77724249eca12eeade1928b7d08412a772d50964de16a").should be_valid
>    end
>  end

pachonk:
Close, but you should split each assertion into it's own `it`. Also, use `expect` matchers.

troystarwalt:
I've pushed the changes and removed the errant inventories_users. Please double check that the schema looks good still. It seems find on my end.

spec/models/user_spec.rb
```diff
@@ -5,6 +5,16 @@
...
+  describe User do
+   it "validates password length" do
+     expect(FactoryGirl.build(:user, password: "1234567")).not_to be_valid
+   end
+
+   it "validates password length" do
+     expect(FactoryGirl.build(:user, password: "a42d1edd2637268dbd053521828ed3eecd5e9e2e95bea57a14d5fe453ba1399c66854004a3d7e13b16d77724249eca12eeade1928b7d08412a772d50964de16a")).to be_valid
+   end
+ end
+
```
Exellin: Try to put both of these into a describe '#password' block.  Right now the tests say 'User validates password length'.  It kind of makes sense, but there shouldn't be 2 tests with the same name.  The test should make it so that if someone changes the password length in the future, it shows an error.  So 1 test right below the password length should be valid, and 1 above should be invalid.

Take a look at describing name and description here.
https://github.com/emerald-rubies/tenacious/blob/master/spec/models/inventory_spec.rb

troystarwalt: Okay - it makes more sense when the example provided. I've done very little testing, so this is all very helpful for me. Appreciate you taking the time to take a look at it.

Exellin:
You can tell the Schema is fine now because when you look at the difference in files the Schema no longer shows up.  You changed it by resetting the database and running migrations though instead of manually editing the Schema right?  Otherwise your database will be incorrect.

troystarwalt:
Yeah, at first I just removed it manually and ran db:migrate. It put it back in. (Which is odd as I didn't see the migration.) Then I did run bin/setup and it took care of it.

pachonk:
@troystarwalt Looks like travis build failed.

troystarwalt:
@pachonk Thank you. Didn't realize that. I'll take a look at it and fix my errors. (Looks to be formatting issues.) Sure keeps you honest.

Exellin:
@troystarwalt.  If you run rubocop and rspec with no errors travis should pass.

Also you still need to update your tests look like before this is ready to merge.

troystarwalt:
I just noticed that there were some additional tests added since I pulled in master. I'll need to update some tests.

spec/models/user_spec.rb
```diff
@@ -5,6 +5,20 @@
     expect(FactoryGirl.build(:user)).to be_valid
   end
 
+  describe '#password' do
```
pachonk: More like `@password` than `#password`.

Exellin: http://www.betterspecs.org/#describe

Betterspecs says to use . for class methods and # for class method name (I think that means attribute).  We used # for all other attribute tests.

pachonk: Just an idea. # is like "method" and really, password isn't a method. It's fine as is though. Does need to drop bad commits from merging though.

troystarwalt: Is that something i need to do? If so, I could use some direction on that.

Exellin: Yeah you need to fix the commit history of this branch since there are 19 commits.  Are you familiar with rebasing?  I can probably try to help you if you don't know where to start.

troystarwalt: @Exellin I've used rebase before but didn't here as I read that you arent supposed to rebase on branches that others will see. And I've never done it after I have pushed. How would you recommend moving forward?

Exellin: Hmm not sure how/why you heard that.  Alex taught me rebasing for this project and it has been working out great.

Anyway I think you have 19 commits since you are based on a commit earlier than the latest version of master, therefore it includes all of the commits from there to master to catch up, and then starts on your commits

I think if you rebase on the current version of master, those commits will be dropped,

I follow the following commands

```
- git fetch upstream
- git checkout upstream/master
- git checkout your-branch-name
- git rebase -i upstream/master (this will open a terminal.  You can change the letters to the left of the commit to things like f for a fixup commit (it will join the commit above it and delete the commit message), or d to delete the commit)
For your first round through you can just save the file and it will attempt the rebase.  You might need to resolve merge conflicts
- Use git mergetool to resolve merge conflitcs, then git add filename or git add -A to put them back into the rebase
- git rebase --continue once all merge conflicts are resolved
- git push -f once the rebase is successfull.  
```

You might also need to delete the extra commits from the rebase file if rebasing onto master doesn't work.  I'm not too familiar with the details of rebasing yet.

Let me know if this information is correct @pachonk

pachonk: Minus the git add -A, that can lead to artifacts accidentally being added, it's good.
I recommend not using the tacks, and adding `# comments` on the lines that can't execute.

vamsipavanmahesh:
I actually did a little digging on @Exellin comment on assert `password_length` and went ahead and wrote this `expect(User.password_length).to be_between(11, 256)` but .. but there is a catch for `be_between` to work User should be including comparable module as per docs, but it's throwing an error does not respond to >= and <=  .
 
So we can use below these two

`User.password_length.min, User.password_length.max` to assert as per the spec :)

Issue #57, opened 2017-07-05
As a user, I want to be able to have a centralized location to see all inventories and orgs I'm associated with


pachonk:

pachonk:
@Exellin Comma separated doesn't work from https://github.com/emerald-rubies/tenacious/pull/96

Issue #33, opened 2017-07-03
As a freshly authenticated user, I want to see a dashboard that shows my my orgs, my inventories, and buttons to add ether/or


pachonk:

Exellin:
This can be closed now.

PR #96, opened 2017-07-29
Create Inventory new, create, and show actions with polymorphic associations

### LINKED ISSUE #79: As a user, I want to be able to add an inventory and associate it with myself or any of my organizations

Exellin:
resolves #79, #57, #33 

Create polymorphic association between inventories and an owner which can be a user or an organization

inventory actions are nested inside of organization and user routes.  Might want to create a user show view or profile so that users/user_id is a valid route in the future.

Lots of improvement to be done on dashboard, inventory etc but this pull request was getting large enough as is and can be done in a future commit.

I removed the min 6 characters helpers on password since it didn't stay formatted correctly on error and it is inconsistent with other forms (we don't have required next to fields saying they are required, we don't have a statement next to description saying it can't be more than 255 characters, etc.  We should talk later about client side validations/helpers)

## Basic Dashboard Page
image

## Basic Organization Page
image

## Form to create an inventory for an organization
image

## Form to create an inventory for yourself
image
Might have a link to username or profile link if we implement that in the future.  Not much point in saying creating inventory for user@example.com at the moment

## Sign up form showing error colors on input fields.
image

### DIFF
--- a/spec/factories/inventories.rb
+++ b/spec/factories/inventories.rb
@@ -2,5 +2,6 @@ FactoryGirl.define do
   factory :inventory do
     name { Faker::Name.unique.first_name }
     description { Faker::Lorem.sentence }
+    association :owner, factory: [:user, :organization].sample
   end
 end
--- a/spec/features/inventories/show_inventory_spec.rb
+++ b/spec/features/inventories/show_inventory_spec.rb
@@ -1,0 +1,69 @@
+require 'rails_helper'
+require 'features/feature_helpers'
+include FeatureHelpers
+
+RSpec.feature 'Showing an inventory' do
+  let(:member) { FactoryGirl.create(:user, :confirmed) }
+  let(:user) { FactoryGirl.create(:user, :confirmed) }
+  let(:inventory) { FactoryGirl.create(:inventory, users: [member]) }
+
+  shared_examples 'has inventory information' do
+    scenario 'shows the name and description of the inventory' do
+      expect(page).to have_content(inventory.name)
+      expect(page).to have_content(inventory.description)
+    end
+  end
+
+  feature 'as a member of the inventory' do
+    before do
+      login_as(member)
+    end
+
+    feature 'belonging to an organization' do
+      let(:organization) { FactoryGirl.create(:organization) }
+      let(:inventory) { FactoryGirl.create(:inventory, owner: organization, users: [member]) }
+
+      before do
+        visit organization_inventory_path(organization, inventory)
+      end
+
+      include_examples 'has inventory information'
+
+      scenario 'has a link to the organization' do
+        expect(page).to have_content(organization.name)
+      end
+    end
+
+    feature 'belonging to a user' do
+      before do
+        inventory.owner = member
+        visit user_inventory_path(member, inventory)
+      end
+
+      include_examples 'has inventory information'
+    end
+  end
+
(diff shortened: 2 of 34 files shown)


spec/models/inventory_spec.rb
```diff
@@ -54,4 +54,22 @@
...
+
+  describe '#owner' do
+    it 'is invalid when nil' do
+      expect(FactoryGirl.build(:inventory, owner: nil)).not_to be_valid
+    end
+
+    it 'can be a user' do
+      user = FactoryGirl.create(:user)
+      inventory = FactoryGirl.create(:inventory, owner: user)
+      expect(inventory.owner_type).to eq('User')
```
pachonk: Consider making these lowercase so you can use `:symbols` instead.

spec/models/inventory_spec.rb
```diff
@@ -54,4 +54,22 @@
...
     end
   end
+
+  describe '#owner' do
+    it 'is invalid when nil' do
+      expect(FactoryGirl.build(:inventory, owner: nil)).not_to be_valid
+    end
+
+    it 'can be a user' do
+      user = FactoryGirl.create(:user)
```
pachonk: Can you use build_stubbed_ on these two and still make it work?

pachonk: As in, test passes?

spec/features/inventories/show_inventory_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      expect(page).to have_content('You need to be part of this inventory to view it')
+    end
+  end
+
+  feature 'as a guest' do
+    before do
+      visit user_inventory_path(user, inventory)
+    end
+
+    will_redirect_to_login_page
```
pachonk: Consider making this `will_perform_unauthenticated_action` so if/when we later change the functionality of non-logged in users, we can change it within the function. This way, if we changed the contents of will_redirect_to_login_page, it would technically be a misnomer. Otherwise, you'd be just as well off asserting it directly without the helper definition

spec/factories/inventories.rb
```diff
@@ -2,5 +2,6 @@
   factory :inventory do
     name { Faker::Name.unique.first_name }
     description { Faker::Lorem.sentence }
+    association :owner, factory: [:user, :organization].sample
   end
```
pachonk: I would make traits to force it to have an owner of user or org as well, probably.

Exellin: Traits are unnecessary.  If you look throughout all of the tests you will see that I often use owner: user or owner: organization.  This factory is just to give the inventory an owner so that it is valid when testing other aspects of the inventory.

pachonk: :+1:

spec/models/inventory_spec.rb
Exellin: By default the owner type is the string of the class name.  I never set it to be uppercase.  If it set it to lowercase for symbols would I have to go through something like this?

https://www.quora.com/Rails-3-1-for-a-polymorphic-association-is-there-a-way-to-change-the-string-looked-up-in-the-owning-sides-type-column

Seems like a lot of work for no real benefit unless you can convince me otherwise.

PR #99, opened 2017-07-31
Admin create user inventory read, write, and admin associations

Exellin:
Adding the username model for possible review while I work on the association.


pachonk:
What happens to users that currently don't have a username?

Exellin:
I just reset the database before I ran this migration.  I would also reset the heroku database.  I didn't think this would be an issue until we have real users on our site.  Or am I wrong here?

pachonk:
More of a hypothetical thing. We can either make a deal out of it and intentionally provide a migration path... Or we can reset the db.

Exellin:
I'm cool with just resetting the database for now.  This probably won't be the last time we add a required field to an existing table.

Exellin:
Pushed everything I have at the moment with the feature working.  The only thing that is missing is updating the show dashboard spec to show pending inventories, and the show inventory spec for showing the users sorted by their access.

app/views/inventories/usersearch.html.erb
```diff
@@ -0,0 +1,13 @@
+<h2>Invite a user to <%= @inventory.name %></h2>
+
+<% @users.each do |user| %>
+  <% unless @inventory.users.include? user %>
```
pachonk: Instead of implementing this check, simply load ONLY the users associated with this. Looks like `@users = @inventory.users` might work?

app/views/inventories/show.html.erb
```diff
@@ -4,3 +4,51 @@
 <% if @inventory.owner_type == 'Organization' %>
   <%= link_to @inventory.owner.name, organization_path(@inventory.owner) %>
 <% end %>
+
+<div class='card'>
+
+  <% unless @admins.empty? %>
```
pachonk: Put all the users in the same table and add a column (to the markup) "role" I'd say. Also, prevents having to query each piece individually.

app/helpers/application_helper.rb
```diff
@@ -3,4 +3,8 @@ def render_errors_for(obj)
     return unless obj.errors.any?
     render 'layouts/errors', obj: obj
   end
+
+  def find_inventory_user(inventory, user)
+    InventoryUser.where(user_id: user, inventory_id: inventory)[0]
```
pachonk: Use .first

app/controllers/pages_controller.rb
```diff
@@ -6,6 +6,14 @@ def home
 
   def dashboard
     @organizations = current_user.organizations
-    @inventories = current_user.inventories
+    @pending_inventories = []
```
pachonk: Try to write these as sql queries so you don't have to prematurely iterate in rb

app/controllers/inventory_users_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+    inventory_user.confirmed_at = Time.now
+    inventory_user.save
+    flash[:success] = 'You have successfully joined this inventory'
+    redirect_to [@inventory.owner, @inventory]
+  end
+
+  private
+
+  def inventory_user_params
+    params.permit(:user_id, :inventory_id, :user_role)
```
pachonk: No require?

app/controllers/inventory_users_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def find_inventory
```
pachonk: I usually call this load_inventory

app/controllers/concerns/inventory_access.rb
```diff
@@ -0,0 +1,10 @@
+# This module adds methods to check for inventory permissions between controllers
+module InventoryAccess
+  private
+
+  def require_inventory_membership
```
pachonk: I would scope the find of the inventory to the current_user's inventory and let AR handle the 404. (404 is more secure because then the user can't sniff the data of inventories and such)

pachonk: `current_user.inventories.find(asdf)`

app/helpers/application_helper.rb
Exellin: done and also changed in all other places [0] was used.

pachonk: Btw, this both 1) exists in Ruby and 2) ONLY fetches the FIRST record from the DB if > 1 exists.

pachonk: The second thing is an AR thing though

app/controllers/inventory_users_controller.rb
Exellin: I'm not actually sending a inventory_user object here so I can't require it.

pachonk: Can you link me to the form?

app/views/inventories/usersearch.html.erb
```diff
@@ -0,0 +1,13 @@
...
+    <div class='card'>
+      <h4><%= user.username %></h4>
+      <%= form_tag(inventory_users_path(user_id: user, inventory_id: @inventory)) do %>
```
pachonk: Use `form_for` here to be able to .require

app/controllers/inventory_users_controller.rb
Exellin: changed.

app/controllers/inventory_users_controller.rb
Exellin: app/views/inventories/usersearch.html.erbhttps://github.com/Exellin/tenacious/blob/033821595f2ac69bfe2639ae304598206293f636/app/views/inventories/usersearch.html.erb

app/views/inventories/usersearch.html.erb
Exellin: I guess I should probably just move the usersearch action from the inventories controller to inventory users controller since I would have to call InventoryUser.new to create the instance variable for this form.  Maybe even just rename the action to new within the controller.

pachonk:
I think that's what I would do personally

app/views/inventories/usersearch.html.erb
Exellin: done

app/controllers/concerns/inventory_access.rb
Exellin: Do I need to figure out how to include rescue_from in this module so I can rescue from RecordNotFound with the flash message and redirect?  Not sure how else to approach this.

pachonk:
Rails rescues with a 404

app/controllers/concerns/inventory_access.rb
Exellin: Ok implemented this change.

app/controllers/concerns/inventory_access.rb
```diff
@@ -0,0 +1,12 @@
+# This module adds methods to check for inventory permissions between controllers
+module InventoryAccess
+  private
+
+  def load_inventory
+    @inventory = Inventory.find(params[:id])
+  end
+
+  def require_inventory_membership
+    return if current_user.inventories.find(@inventory.id)
```
pachonk: Remove this whole function and use `current_user.inventories.find(@inventory.id)` to load the inventory.

app/controllers/concerns/inventory_access.rb
```diff
@@ -0,0 +1,12 @@
+# This module adds methods to check for inventory permissions between controllers
+module InventoryAccess
```
pachonk: I'm actually all about being explicit and including all these setters in the class I'm currently in.

app/controllers/concerns/inventory_access.rb
```diff
@@ -0,0 +1,12 @@
...
+  private
+
+  def load_inventory
```
pachonk: I actually learned recently that this is almost always a bad idea and you should inline this loader to be more explicit.

pachonk: @arthurthefourth pointed me in the right direction here for sure.

app/controllers/pages_controller.rb
Exellin: done

app/views/inventories/show.html.erb
Exellin: Implemented the table.  Unless you meant not having to check if the array was empty for displaying the header for the roles, I'm not sure how to avoid querying each group of users (admin, write, read, invited) individually.  I can't just check for the user role since they need to be confirmed as well.  This was the purpose of the inventory_access? method in user.rb

I also can't just check if the user has inventory_access, since admins will return true for all roles etc.  So I created the admin, writers, readers, functions which go through the inventory_access method checking for both the user_role and a confirmed value.  Let me know if you can see an easier way to do this.

app/views/inventories/usersearch.html.erb
Exellin: I'm not sure what you mean here.  Right now @users is the result of the search for users to invite to the inventory.  This check doesn't display the user if they are already invited or part of the inventory.

app/views/inventories/show.html.erb
Exellin: I guess I could make a query to find all confirmed users, and then use the helper method I made find_inventory_user to find the user role for each user like I did to list the role for the invited users.

app/controllers/concerns/inventory_access.rb
Exellin: Needed to use @inventory = current_user.inventories.find(params[:inventory_id]) since @inventory isn't loaded yet, but I really like this change since it removes one more check in the controller.  Thanks for the suggestion.

app/controllers/concerns/inventory_access.rb
Exellin: done

app/controllers/concerns/inventory_access.rb
Exellin: Yeah this goes back to my obsession with making the controllers DRY which I understand now is not always something to strive for.  With the removal of the require_inventory_membership function this would be the only function in the module.  I put them back in their respective controllers since there is only 1 duplication and it makes the code easier to browse through.

This also meant I could change back the inventory_access id to inventory_id since it is slightly more transparent what the id is talking about for the relationship and I deleted the entire module.

Exellin:
More of a Todo for me/question:  I realized I am also sending hte user_id as a parameter for the inventory_users_controller and everything works without it being permitted since it isn't in the form.

For the user_id and inventory_id, is it better to send them over as parameters in the url or as hidden fields in the form?  Which is more secure to prevent manipulation?

pachonk:
For manipulation, neither can be deemed secure.

Exellin:
I'm removing the WIP tag from this PR.  All features have been implemented and tested sufficiently in my opinion for this to be a complete PR.  All other improvements (UI, creating controller tests, optimizing database queries can either be requested in this PR or added in future PRs.)

spec/views/pages/dashboard.html.erb_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'rails_helper'
+
+RSpec.describe 'pages/dashboard', type: :view do
+  let(:user) { FactoryGirl.build(:user, :confirmed) }
+  let(:owner) { FactoryGirl.build(:user) }
+  let(:organization) { FactoryGirl.build(:organization, users: [user]) }
+  let(:organization_inventory) { FactoryGirl.build(:inventory, owner: organization) }
+  let(:user_inventory) { FactoryGirl.build(:inventory, owner: owner) }
+  let!(:pending_inventory) { FactoryGirl.create(:inventory, owner: owner, users: [user]) }
```
pachonk: Why is your view hitting the DB at all?

spec/views/pages/dashboard.html.erb_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'rails_helper'
+
+RSpec.describe 'pages/dashboard', type: :view do
+  let(:user) { FactoryGirl.build(:user, :confirmed) }
```
pachonk: build_stubbed instead.

spec/views/pages/dashboard.html.erb_spec.rb
Exellin: For this view I can't figure out how to create the inventory_user relationship with a stubbed model.  

[This is the closest stackoverflow question I could find related to my issue](https://stackoverflow.com/questions/17754770/factorygirl-build-stubbed-strategy-with-a-has-many-association)

Would you be able to quickly browse through it and see if I should explore some of the solutions proposed there?  Or is there an easier solution, or should I stick with creating the models?

pachonk:
At dinner. Will look later, remind.

pachonk:
Try making a trait with_inventory to make the inventory within the user perhaps?

app/models/inventory.rb
```diff
@@ -37,4 +37,8 @@ def invited_users
     end
     invited_users
   end
+
+  scope :confirmed, -> { where.not(inventory_users: { confirmed_at: nil }) }
+
```
pachonk: I would put scopes close to the top. After validations and before relations usually.

pachonk: Also, I would "push the lines together" just like your validations and relations

app/models/inventory.rb
```diff
@@ -37,4 +37,8 @@ def invited_users
...
+  scope :pending, -> { where(inventory_users: { confirmed_at: nil }) }
```
pachonk: Ha, you could probably make a `confirmed` scope on inventory users and simplify this `confirmed_at: nil` to make more sense and make it readable.

Exellin:
Just a quick note to correct me if I'm wrong.  I don't think there is any value in writing tests for the confirmed and pending scopes in the inventory model.

This is because Inventory.confirmed and Inventory.pending don't really have any meaningful value without a user.  An inventory can be in the inventory_users table multiple times with entries being confirmed or pending depending on the relationship with the user which is part of that inventory.

The only time we care about these scopes is for user.inventories and user.pending_inventories, which I already wrote tests for in the user spec.

app/models/inventory.rb
Exellin: done

app/models/inventory.rb
Exellin: I don't really think there is anything unreadable here.  It would just be 1 more model a developer would need to look through to see what scope is really looking at.

Exellin:
All view specs now done with stubbed models.  2 things at the moment I am not sure how to do is show the invites users role without hitting the database so I removed it.  The other is how to test a confirmed users role is in the correct row on the table.

app/models/inventory.rb
pachonk: See, that's _actually_ the point that I'm trying to avoid. This model doesn't need to know when an inventory_user is confirmed, it only needs to know "I want all confirmed inventory_users".
Separation of concern is the reason to make this a confirmed scope.

Exellin: This model is outdated I moved the sql query to the inventory_user model.

pachonk: All good, just explaining the justification

app/views/inventories/show.html.erb
```diff
@@ -4,3 +4,55 @@
...
   <%= link_to @inventory.owner.name, organization_path(@inventory.owner) %>
 <% end %>
+
+<div class='card'>
+<h4>Users with access to this inventory:</h4>
+
+  <table>
+    <thead>
+      <tr>
+        <th>Username</th>
```
sanchez2210: I think we should identify users by their name instead!

app/views/inventories/show.html.erb
```diff
@@ -4,3 +4,55 @@
...
+        <li><%= "#{user.username}" %></li>
+      <% end %>
+    </ul>
+  <% end %>
+
+  <% if current_user.inventory_access?('admin', @inventory) %>
+    <h4>Search for a user to add to this inventory:</h4>
+
+    <%= form_tag(new_inventory_user_path(@inventory), method: :get) do %>
+      <%= text_field_tag :query, nil, placeholder: 'part of username or email' %>
```
sanchez2210: Do we need the 'part of' bit?

app/views/inventory_users/new.html.erb
```diff
@@ -0,0 +1,13 @@
+<h2>Invite a user to <%= @inventory.name %></h2>
+
+<% @users.each do |user| %>
+  <% unless @inventory.users.include? user %>
+    <div class='card'>
+      <h4><%= user.username %></h4>
```
sanchez2210: Same here, I'd say full name instead of username.

pachonk:
I'd rather do something like:
Alex Piechowski (@pachonk)

app/views/inventories/show.html.erb
Exellin: I agree since this will mostly be used in a professional environment that a name would be more appropriate than a username.  I would change it to first_name and last_name with a full_name method to tie them together.

pachonk:
Sorry, what I was suggesting is that we still show the username in the name
column, but in parentheses.

Exellin:
If we decide to add a name to users and keep username, then it can be added in a future PR.

pachonk:
Maybe I misunderstood his original request. Let me look this whole PR over
as is when I get back and we'll go from there.

app/views/inventories/show.html.erb
sanchez2210: Yes! You're right though, we can add the name attributes in a later PR

app/controllers/inventories_controller.rb
```diff
@@ -20,15 +21,14 @@ def create
...
-    return if @inventory.users.include? current_user
-    flash[:alert] = 'You need to be part of this inventory to view it'
-    redirect_to root_path
-  end
+  def show; end
 
   private
 
+  def load_inventory
+    @inventory = current_user.inventories.find(params[:id])
```
pachonk: I would much prefer not using a hook to load ivars

app/controllers/inventory_users_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+      flash[:alert] = 'You must be an administrator of this inventory to view this page'
+      redirect_to [@inventory.owner, @inventory]
+    end
+    @users = User.search(params[:query])
+    @inventory_user = InventoryUser.new
+  end
+
+  def create
+    @inventory_user = InventoryUser.new(inventory_user_params)
+    @inventory_user.user_id = params[:user_id]
```
pachonk: Why isn't user_id just in the strong params?

spec/factories/inventories.rb
```diff
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :inventory do
-    name { Faker::Name.unique.first_name }
+    sequence(:name) { |n| "#{Faker::Name.unique.first_name}#{n}" }
```
pachonk: Why sequence and faker unique?

pachonk:
I really want to avoid PR's from getting this big in the future, that's a hard review to look through.

app/models/user.rb
```diff
@@ -5,9 +5,31 @@ class User < ApplicationRecord
...
+  has_many :pending_inventories, -> { pending }, source: :inventory, through: :inventory_users
   has_many :owned_inventories, class_name: 'Inventory', as: :owner
   has_many :owned_organizations, class_name: 'Organization', foreign_key: 'owner_id'
+
+  def self.search(query)
+    where('username ilike :q or email ilike :q', q: "%#{query}%").limit(5)
+  end
+
+  def inventory_access?(user_role, inventory)
+    permission = 'admin' if inventory.admins.include?(self)
```
pachonk: I feel that we should just ask what the user's role is rather than doing this piece, actually.

spec/factories/inventories.rb
Exellin: There are not enough unique faker names and I was starting to get duplicates in my tests.

pachonk:
Okay, that will actually throw an exception as is if you run out of faker
names.. You should probably remove the unique.
Also, if we're running out of names that shows a code smell that our tests
are making too many users.

app/models/user.rb
Exellin: Just asking what the user's role is doesn't take into account a user having lower level permission.  Let's say if you were asking if a user had write access just asking if it was part of writers would lead to admins not having write access.

app/controllers/inventory_users_controller.rb
Exellin: user_id and inventory_id are in the params of the route for the form

```
      <%= form_for @inventory_user, url: inventory_users_path(user_id: user, inventory_id: @inventory) do |f| %>
        <%= f.select(:user_role, [['read'], ['write'], ['admin']]) %>
        <%= f.submit 'Invite User' %>
      <% end %>
```

Adding them to strong params would mean making hidden fields with the user_id and inventory_id.  I asked earlier which one was more secure on August 15 in this PR and you said neither so I stuck with the method I am using now.

app/controllers/inventories_controller.rb
Exellin: What does a hook mean in this context?

app/controllers/inventory_users_controller.rb
pachonk: Ahh, okay okay.

app/models/user.rb
pachonk: `user.inventory_users.where(inventory: inventory).user_role` or similar should accomplish this piece, right?

Exellin:
Removed the unique for all factories.  Not sure how to approach the code smell.  To be fair this was after running thousands of tests over weeks and not resetting the database.  I can't see it ever reaching the limit in a single run of the test suite.

pachonk:
Hmm, the tests should be running transactionally... Wait! I think capy
tests don't run in transactions, come to think of it!
We should implement database cleaner then, perhaps.

app/models/user.rb
pachonk: Unless I'm also misunderstanding.

sanchez2210: This is great candidate of an InventoryPolicy. With the gem we were talking about, Pundit.

Exellin: `user.inventory_users.where(inventory: inventory).user_role` doesn't take into account whether it is confirmed or not.  Which is why I went through the inventory queries.  There may be a way to easily query the user_role which is also confirmed which I am not seeing.

pachonk: The solution would be to toss the confirmed into the where clause then, or better yet make the confirmed scope.

Exellin: I got it working with the following:

```
    inventory_user = inventory_users.confirmed.where(inventory: inventory).first
    return unless inventory_user
    permission = inventory_user.user_role
```

However now the inventory view spec is failing since there is no inventory_user relation set up between the user and the inventory.

sanchez2210: I think the problem is that you're calling this from @user (current_user).
I understand that you in non technical terms your question is: "Does this user has access to this inventory?" but thinking with code it's the InventoryUser who needs to know if it has enough permissions.
```
<% if @inventory_user.role == 'admin' %>
  ...
<% end %>
```

PR #101, opened 2017-08-04
Update README.md

jeison24:


pachonk:
I agree with @Exellin for the reason that where it's imported, none of the other css files have the font to use, IIRC.
Also, please stay true to the PR name. The PR name is "Update README.md" but this also adds font. This should be 2 PRs, "Update README.md" and "Chnage font to ___"

pachonk:
Furthermore, take a look at https://chris.beams.io/posts/git-commit/
"Update REAME.md" is pretty bland and can mean many different things. "Fix webpack typo in README.md" is a lot more particular.

jeison24:
Ready @pachonk this PR is only for README file.

PR #109, opened 2017-08-09
Convert show org feature specs to unit specs

pachonk:


Exellin:
What does the .swo file do?  Does it get compiled?  Should it be added to .gitignore?

spec/views/organizations/show.html.erb_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    assign(:inventories, inventories)
+    allow(view).to receive(:current_user).and_return(user)
+    render
+  end
+
+  it "displays the organization's name" do
+    expect(rendered).to include organization.name
+  end
+
+  it "displays a link to the organization's inventories" do
```
Exellin: Should make these quotes single to stay consistent with all other specs.

pachonk: They are double because I have a `'` in the string.

pachonk: Another option is to have a `\'` instead, but I find that a little messy. What's your take?

Exellin: oh my bad didn't catch that

Exellin: I like the double quotes instead.

pachonk: Yeah, and rubocop doesn't whine so I'd say I'm good.

pachonk:
@Exellin interesting that it says the cov went up

pachonk:
.swo was accident. Sorry!

Exellin:
I'm not really sure what exactly coverage means.

pachonk:
Me either.

Exellin:
I think you mentioned you could test all of the features in show_organization_spec inside the view spec.  If so, I think it would be best to finish testing everything inside the view spec before pulling the PR in.

pachonk:
I'm more about incremental changes to prevent PR's from going stale. While
I'll continue working on converting it to unit level specs... I'd rather
get this in first.

spec/views/organizations/show.html.erb_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    expect(rendered).to have_link(inventories.first.name)
```
Exellin: Why create a list of inventories if you are only testing if the first link is shown?

pachonk:
I converted what I could to be view specs, the rest will have to be asserted at the controller level.

spec/views/organizations/show.html.erb_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'rails_helper'
+
+RSpec.describe 'organizations/show', type: :view do
+  let(:owner) { FactoryGirl.build_stubbed(:user) }
+  let(:member) { FactoryGirl.build_stubbed(:user) }
+  let(:organization) { FactoryGirl.build_stubbed(:organization) }
+  let(:organization) { FactoryGirl.build_stubbed(:organization, owner: owner, users: [owner, member]) }
```
Exellin: You have 2 variables with the same name here.  Can probably delete the first one.

pachonk: Whoops, great catch.

pachonk: Fixed.

spec/views/organizations/show.html.erb_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+RSpec.describe 'organizations/show', type: :view do
+  let(:owner) { FactoryGirl.build_stubbed(:user) }
+  let(:member) { FactoryGirl.build_stubbed(:user) }
+  let(:organization) { FactoryGirl.build_stubbed(:organization, owner: owner, users: [owner, member]) }
+  let(:inventories) { FactoryGirl.build_stubbed_list(:inventory, 2, owner: organization) }
+
+  before do
+    assign(:organization, organization)
+    assign(:inventories, inventories)
```
Exellin: This might be going past what a view spec is supposed to be testing (and partly why I am having trouble creating stubbed models for my own tests which aren't supposed to touch a database).  But assigning inventories to inventories instead of organization.owned_inventories has no assurance that the inventories in the view are actually owned by the organization.

To demonstrate this point, you can remove the owner parameter in the inventories assignment and the test still has everything pass.  Using organization.owned_inventories also fails the test as the inventories aren't actually owned by the organization in the database (since it isn't touching one).

Let me know what your thoughts are here and whether this doesn't matter for a view test.  But then I am also questioning the value of a stubbed model if it isn't accurately portraying the relationships in the test.

pachonk:
Hmm, I get what you're saying. That implementation detail is more of a
controller level piece so I'll do that before this PR is pulled in. Fair?

Exellin:
Yeah that would be great.  However I am still uncomfortable having the test say 'displays a link to the organizations inventories' with the stubbed models when they aren't the organizations inventories in this test.  It would be much more true to what we are testing here if it said 'lists the inventories instance variable'

pachonk:
How are they "not the organization's"?

Exellin:
organization.owned_inventories returns an empty array for this test.

pachonk:
Is that utilized within the view?

Exellin:
@inventories is assigned to organization.owned_inventories in the controller which handles this view.

pachonk:
In that case, my opinion is that the view spec can "reap context from the
controller" in the description. Even though we're hard assigning
inventories to a variable. I think it's less informational the intention of
"it displays the given inventories" even though that's technically it's
job. /cc @brandondees @arthurthefourth

Exellin:
By "reap context from the controller" does that mean we will test that the inventories instance variable will be assigned properly there, and therefore it can be hard coded inside the view spec since it was already tested there?

I also don't know what you mean by "I think it's less informational the intention of"

pachonk:
What I'm saying is "yes I agree with the fact that the view's
responsibility is displaying the instance variable as links" but it is more
imformational to people reading our documentation to State the view is
displaying the org's inventories

pachonk:
I think I want to get a second or third opinion on this. I see strong arguments for both ways.

pachonk:
@Exellin PR moved to WIP until controller tests added.

pachonk:
@Exellin check it out

spec/controllers/organizations_controller_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    context 'when user is a part of the organization' do
+      before do
+        allow(subject).to receive(:current_user).and_return(user)
+      end
+
+      it 'assigns @organization' do
+        subject.show
+        expect(subject.view_assigns['organization']).to eq organization
```
Exellin: We have been surrounding all parameters with brackets for all of our tests.  Rubocop doesn't test for it but good to be consistent.

spec/controllers/organizations_controller_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      end
+
+      it 'assigns @inventories' do
+        subject.show
+        expect(subject.view_assigns['inventories']).to match_array(inventories)
```
Exellin: Check out the failed travis test for my PR for user roles.  A test that had been running successfully for probably hundreds of iterations suddenly failed because the array got out of order somehow.  My solution was to add .sort onto the array.  Let me know if you think that is reasonable, and if so implement it here as well.

spec/controllers/organizations_controller_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'rails_helper'
+
+RSpec.describe OrganizationsController do
```
Exellin: Should declare the type of test this is.

spec/controllers/organizations_controller_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  let(:user) { FactoryGirl.create(:user) }
```
Exellin: The user only has to be built and not created.  This would save one more database hit.

spec/controllers/organizations_controller_spec.rb
pachonk: match_array should account for that.

spec/controllers/organizations_controller_spec.rb
pachonk: Nope, it's no "type" of test. It's a PORO test.

spec/controllers/organizations_controller_spec.rb
pachonk: Hmm, I disagree with ALWAYS requiring parens. It's kind of a Ruby right of passage. Would you like to have further discussion?

spec/controllers/organizations_controller_spec.rb
pachonk: It will be created regardless from the let under this

spec/controllers/organizations_controller_spec.rb
pachonk: Again, not trying to be authoritative here, it's just a massive "yay ruby" feature.

spec/controllers/organizations_controller_spec.rb
Exellin: not even type: :controller?  What is the purpose of type: :model or type: :view?

pachonk: This is testing my controller but I don't need all the stupid rails additions for controller specs.

pachonk: The type installs extra stuff, rspec matchers, etc

spec/controllers/organizations_controller_spec.rb
Exellin: ok fixed my commit to all just use match_array.  Thanks

spec/controllers/organizations_controller_spec.rb
Exellin: I'm not trying to discuss here whether or not to use brackets, just whether you should use brackets here to be consistent with the rest of the repository where brackets have been used for the eq function.

To quickly discuss use of brackets in ruby quickly, I basically always use them but I heard a good rule of thumb which I would be ok with implementing is only using brackets when there are 2 or more parameters.

Is there a general rule you follow?

spec/controllers/organizations_controller_spec.rb
Exellin: I just looked in rails_helper and found this segment:

```
  # RSpec Rails can automatically mix in different behaviours to your tests
  # based on their file location, for example enabling you to call `get` and
  # `post` in specs under `spec/controllers`.
  #
  # You can disable this behaviour by removing the line below, and instead
  # explicitly tag your specs with their type, e.g.:
  #
  #     RSpec.describe UsersController, :type => :controller do
  #       # ...
  #     end
  #
  # The different available types are documented in the features, such as in
  # https://relishapp.com/rspec/rspec-rails/docs
  config.infer_spec_type_from_file_location!
```

So right now the type is being inferred from the location and all tests are being loaded with the extra rspec matchers.  All of the tests work without specifying the type.  I forget if you preferred to state the type for semantics or not.  Should we comment out this line and only state the type when needed, keep the line and specify all types for semantics, or keep the line and remove all types from our tests?

pachonk: Dang, I didn't know that. Maybe I should create a different folder for my tests then... controller-unit perhaps?

pachonk: My vote is "Always state the type"

spec/controllers/organizations_controller_spec.rb
pachonk: Brackets = `[]` and parens = `()` to me btw.
There are other cases that parens increase readability or are even required to make things work. If it looks good and understandable without them, I drop them. Here it reads REALLY nicely.
`blah.to eq org` and the parens would just feel like extra clutter to me.

spec/controllers/organizations_controller_spec.rb
Exellin: So if you want to always state the type, I think it would be beneficial to comment out that line in order to keep that ruling supported throughout the repository.  I'm not a fan of code that does nothing when it is fairly obvious the test is a feature test when it is in the feature folder.

pachonk: I don't care I guess. I think there's value in persisting the default settings, but I see your point either way.

spec/controllers/organizations_controller_spec.rb
Exellin: I also just did a quick test with debugger seeing the value of user when it is built and user when it is created.

When built, the user has a different username and email because a brand new user object is created and put into the database when user is loaded when the organization is created.

pachonk: Screenies?

spec/controllers/organizations_controller_spec.rb
Exellin: So therefore it is a case by case basis generally depending on the preference of the developer and it is ok if there are some one parameter functions with parens and some without even within the same repository.

I think I still prefer them here (just due to my background and because vscode doesn'y highlight it as a variable without parens).  But its not such a big deal that I'll force other developers to use them.

spec/controllers/organizations_controller_spec.rb
Exellin: image

Exellin: image

pachonk: And does it work "as expected" with create?

pachonk:
Sweet

spec/controllers/organizations_controller_spec.rb
Exellin: Create makes more sense since if you wanted to do any queries on the user object, it would be the same user object as the one in the database which owns the organization in this test.

pachonk:
Yeah, I agree. Your comment just didn't make sense so it's easier to see
the results than read an explanation like that.

pachonk:
So @Exellin, can I get your :+1: ?

Exellin:
I would still like a 2nd opinion on the show test.

My current issues with it:

The organization and inventory instance variables are currently:
```
  let(:organization) { FactoryGirl.build_stubbed(:organization, owner: owner, users: [owner, member]) }
  let(:inventories) { FactoryGirl.build_stubbed_list(:inventory, 2, owner: organization) }
```

and can be replaced with the following and still pass due to the relationships not being properly modelled with build_stubbed.
```ruby
  let(:organization) { FactoryGirl.build_stubbed(:organization, owner: owner) }
  let(:inventories) { FactoryGirl.build_stubbed_list(:inventory, 2) }
```

Interestingly, the owner parameter is still required and works properly for the test.  Probably because the has_one relationship works properly for stubbed models, and not has_many like in my PR.  I can understand why the users parameter is not needed since getting access to the organization by being a member is on the controller level.

I would also like to see the assignment of inventories change from:
`assign(:inventories, inventories)` to `assign(:inventories, organization.owned_inventories)`

If that is not possible, then I would like to see the name of the test 'displays links to the organization's inventories` change to `displays links to the inventories instance variable`.  Then like in the video you should me for why integration tests are bad with one unit test saying `if the inventories instance variable exists, they are displayed.  Then in another unit test saying `the inventories instance variable is properly assigned to the organizations inventories`.  If both of those pass, then then the inventories instance variable is both properly assigned and displayed and no integration test is needed.

brandondees:
if the relationship between models is something the view is responsible for, you may have code in the wrong layer of the application.

that said, it's an antipattern to have more than one instance variable per controller action for the view to rely on. often it can be minimized by accessing relations of AR objects directly through method chaining, but the more ideal pattern is to craft a presenter object for each complex view to depend on, which provides exactly what is needed with only one ivar, and only one level of method calls needed.

pachonk:
In this case, may I suggest we get this test pulled in and create an issue to refactor the org#show action? (I'll execute that) @Exellin @brandondees

Exellin:
This PR is just changing tests and there are not features which depend on this, so I don't see a rush to pull this in to refactor later.

pachonk:
Basically I'd rather have this be a referable commit of pure testing changes, and I want another commit to refer to "changes that required test changes to finish the task". That's why I would like this pulled in as is.

pachonk:
And I kind of wanted to "show" how the refactor makes the `it` more _true_ to the message. That said, I can understand if the approval is a "hard no" to get this in.

Exellin:
This is my first repo and not sure about about best practices for PRs.  I could see a lot of people refusing to merge "incomplete code" since it would then be very easy to ignore it since it is already part of master.  However since this is more about learning and I can see the value of demonstrating how to refactor this test, I'm ok with approving this PR.

pachonk:
@Exellin I would agree on a "typical" OSS incomplete wouldn't be pulled in. Something this small might just be a "whatever, just pull it in" anyway.
I like the idea of leaving that in there to let an apparent "code smell" exist and provide a refactor PR.

PR #112, opened 2017-08-11
Styles

sanchez2210:


pachonk:
For all ui changes please post screenies

sanchez2210:
image
 That was quick!

pachonk:
Couldn't sleep... Haha
I propose we go full material.io on this.

sanchez2210:
Sounds good, actually that button is the least of the additions. Most of the changes were to make the styles more modular and maintainable.

pachonk:
I'll have to review the code on a PC.

app/javascript/application.scss
```diff
@@ -1,4 +1,7 @@
+@import 'stylesheets/base.scss';
+@import 'stylesheets/layout.scss';
 @import 'stylesheets/custom.scss';
-@import 'stylesheets/navbar.scss';
+@import 'stylesheets/modules/index.scss';
 @import 'stylesheets/alerts.scss';
 @import 'stylesheets/forms.scss';
+@import 'stylesheets/buttons.scss';
```
pachonk: See that little red icon? It signifies that you're missing a new line at EOF. Please add a new line at the end of these files

app/javascript/stylesheets/base.scss
```diff
@@ -0,0 +1,3 @@
+body {
+  font-family: "Lato";
```
pachonk: Is there an issue for changing the font to Lato?

sanchez2210: Oh, I don't know. Looked good. Forgot to add the source though. We could do without the font for now, I'll remove that.

pachonk:
Sounds good.

sanchez2210:
Travis wont build with merge conflicts?

app/views/layouts/_navbar.html.erb
```diff
@@ -1,19 +1,17 @@
...
 
   <% if user_signed_in? %>
-    <ul>
-      <li><%= link_to 'Dashboard', dashboard_path %></li>
-    </ul>
+  <%= link_to 'Dashboard', dashboard_path, class: "navbar__hide" %>
   <% end %>
 
-  <ul>
+  <ul class="nav navbar__nav navbar__hide">
```
Exellin: why add classes to all of these elements?  If you look through [the PR for creating a navbar](https://github.com/emerald-rubies/tenacious/pull/76) we tried to style the navbar with as little classes as possible as it is much easier to maintain and the styles semantically describe the markup.

app/views/pages/dashboard.html.erb
```diff
@@ -29,5 +29,5 @@
     </ul>
   <% end %>
 
-  <%= link_to 'Create an Inventory', new_user_inventory_path(current_user) %>
+  <%= link_to 'Create an Inventory', new_user_inventory_path(current_user), class: "btn" %>
```
Exellin: Quotes should be single here.

app/javascript/stylesheets/modules/index.scss
```diff
@@ -0,0 +1,2 @@
+@import 'nav.scss';
+@import 'navbar.scss';
```
Exellin: Why split this up into 2 files?  Even if you want to have different stlyling for .nav and .navbar (if we want to go back to using classes for styling here), I think they would still benefit from being in the same file since they are highly related to each other.

pachonk: I agree. I've even gone as far to have the header in 1 file if its simple.

app/views/layouts/application.html.erb
```diff
@@ -10,8 +10,10 @@
   <body>
     <%= render 'layouts/navbar' %>
     <main>
-      <%= render 'layouts/flash_messages' %> 
-      <%= yield %>
+      <div class="container">
+        <%= render 'layouts/flash_messages' %> 
+        <%= yield %>
+      </div>
```
Exellin: If you look at my navbar PR you'll see that I had the exact same proposal and we opted to use body > main styling for a cleaner layout with the same purpose.  If we decide we want to extensively use the container class throughout the site it might be worth switching to this layaout then, since it would save redundant styling.  However we should have a discussion about that first.

app/javascript/stylesheets/base.scss
```diff
@@ -0,0 +1,2 @@
+body {
+}
```
Exellin: What do you want to have this file for if it is empty?

app/javascript/stylesheets/layout.scss
```diff
@@ -0,0 +1,4 @@
+.container {
+  max-width: 1100px;
+  margin: 0 auto;
+}
```
Exellin: Do we want to limit the max-width of the site to 1100px since you put this over yield?  Probably worth opening an issue to discuss.

app/javascript/stylesheets/custom.scss
```diff
@@ -3,15 +3,10 @@ body {
...
 
-body > main {
-  padding: 0 20px;
-}
-
 .card {
   padding: 10px 40px;
   border-radius: 4px;
   border: 1px solid #e8e8e8;
-  box-shadow: 0 1px 0 rgba(0,0,0,.25);
```
Exellin: Don't like the shadow here?  Right now at least until we have more of a contrast between the color of the card and the background, I think it helps define the outline more.

app/javascript/stylesheets/layout.scss
Exellin: Also do we need another file here?  I like to keep all miscellaneous stylings in custom.scss unless it is describing a certain component or there are enough stylings in 1 categroy to justify making a file for them.

Exellin:
I like the button stlyings, but I think it would be worth for you to read the [pull request for the navbar](https://github.com/emerald-rubies/tenacious/pull/76) since a lot of the changes you implement here were already discussed in that PR and reversed a lot of the decisions made.  I'm open to them, just have to justify your changes.

app/javascript/stylesheets/modules/index.scss
sanchez2210: Thinking of a navigation element differing from a proper navbar

app/javascript/stylesheets/base.scss
sanchez2210: base styles

sanchez2210:
I see, well I just wanted to style the button and ended up changing many things haha.

app/javascript/stylesheets/base.scss
Exellin: What are those exactly?  We can make the file if needed when we implement a base style.

app/javascript/stylesheets/layout.scss
sanchez2210: I don't think the name 'custom' says much about the content of the file. I also found many instances of body element styles scattered across many files.

app/javascript/stylesheets/base.scss
sanchez2210: I used this to set the base font of the site but I reverted since the font choice is Roboto now. I left the file there with an empty rule as a todo.

app/javascript/stylesheets/custom.scss
sanchez2210: It's a bit too much for my taste

app/javascript/stylesheets/base.scss
pachonk: In that case, let's trash this file. No reason to have an empty style rule.

app/javascript/stylesheets/custom.scss
pachonk: Personal preference is okay to have, but let's not willy nilly remove styles without discussing. If you'd like to propose removing it, let's open an issue to discuss.

app/views/layouts/_navbar.html.erb
sanchez2210: I'm not sure if easier to maintain though. Found some instances of 3+ deep nesting and it wasn't too easy to understand

pachonk:
@sanchez2210 Typically we prefer to split PR's to prevent these kinds of conversations. For instance if we agree with 1 element but NOT the other elements, we can still approve the 1 element. In this case, it's possible we could have accepted the buttons on their own, but there is enough other stuff that needs modifications that this PR is currently at a stand still.
Of course "Whole site makeovers" are a whole different story too.

pachonk:
I propose something like this, but in our color scheme:
image

pachonk:
https://material.io/guidelines/components/buttons.html#buttons-button-types

app/javascript/stylesheets/layout.scss
pachonk: Most similar element style should definitely be grouped together except for certain cases, like we might have a "typography partial" where we target html, body to change the type face.

app/javascript/stylesheets/modules/index.scss
pachonk: I think you're trying to over organize for this stage of the project. For now splitting is YAGNI

app/views/layouts/_navbar.html.erb
pachonk: It is most definitely easier to maintain because if I make a new feature that includes a form elementsI don't need to know ANYTHING about "what classes to use" and "how to organize"... When I code _normal_ HTML it just.... works. You don't have to touch/explore the CSS very often this way :)

sanchez2210: I see. Whenever we want to style two elements the same, we will be doing a good old copy-paste. correct?

pachonk: Can you give an example?

app/javascript/stylesheets/modules/index.scss
sanchez2210: I still do not feel comfortable with the nav element being styled directly. We're locked from using ANY nav in any other part of the site without the fear of unindended styling.

pachonk: Well... Not really actually. AFAIK, there's not other nav element, correct? So it's not a current issue and shouldn't have cognitive energy utilized to think about changing it. That's my opinion anyway.

sanchez2210:
Hi @pachonk @Exellin I've removed most of the modifications of the PR. I've kept the nav_hide because while I do agree that purely stylistic classes might need some discussion, .nav_hide does have some semantic value as well.

PR #125, opened 2017-10-18
Implement a Pundit InventoryPolicy

pachonk:


app/controllers/inventory_users_controller.rb
```diff
@@ -4,15 +4,13 @@ class InventoryUsersController < ApplicationController
   before_action :load_pending_inventory, only: [:confirm_invitation]
 
   def new
-    unless current_user.inventory_access?('admin', @inventory)
-      flash[:alert] = 'You must be an administrator of this inventory to view this page'
-      redirect_to [@inventory.owner, @inventory]
-    end
+    authorize @inventory
```
sanchez2210: Which policy is this line authorizing with?

sanchez2210: I believe this is passing to ApplicationPolicy right? Because there's no `new?` method at InventoryPolicy

pachonk: Should be inventory from the model name.

sanchez2210: There's no `InventoryPolicy#new?` so ApplicationPolicy's `new?` does check for `InventoryPolicy#create?`. Looks good then.

sanchez2210:
Looking good. We've lost the ability to display a custom flash message when not authorized to do something so we might want to create a new issue for that. We can rescue from `Pundit::NotAuthorizedError` at ApplicationController. For now, the need to have ANY authorization structure makes me feel like this PR is very necessary. Good job!

pachonk:
@luis FWIW I'd rather act like a 404 than allow the user to explore the
edges of the system with nice messages honestly

pachonk:
Like literally I sometimes throw a 404 to make it looks like "oh there's no
page here" lol

